PR #5, opened 2017-02-16
fix web module

larat7:
adapts the web module to the new soup api

### DIFF
--- a/src/web/main.rs
+++ b/src/web/main.rs
@@ -1,5 +1,9 @@
 #[cfg(feature="web")]
 extern crate distributary;
+extern crate slog;
+extern crate slog_term;
+
+use slog::DrainExt;
 
 #[cfg(feature="web")]
 fn main() {
@@ -7,18 +11,19 @@ fn main() {
 
     // set up graph
     let mut g = distributary::Blender::new();
+    g.log_with(slog::Logger::root(slog_term::streamer().full().build().fuse(), None));
 
     {
         let mut mig = g.start_migration();
 
         // add article base node
-        let article = mig.add_ingredient("article", &["id", "user", "title", "url"], Base {});
+        let article = mig.add_ingredient("article", &["id", "user", "title", "url"], Base::default());
 
         // add vote base table
-        let vote = mig.add_ingredient("vote", &["user", "id"], Base {});
+        let vote = mig.add_ingredient("vote", &["user", "id"], Base::default());
 
         // add a user account base table
-        mig.add_ingredient("user", &["id", "username", "hash"], Base {});
+        let user = mig.add_ingredient("user", &["id", "username", "hash"], Base::default());
 
         // add vote count
         let vc = mig.add_ingredient("votecount",
@@ -38,9 +43,13 @@ fn main() {
 
         mig.maintain(awvc, 0);
         mig.maintain(karma, 0);
+
+        // commit migration
+        mig.commit();
     }
 
     web::run(g).unwrap();
+    loop {}
 }
 
 #[cfg(not(feature="web"))]
--- a/src/web/mod.rs
+++ b/src/web/mod.rs
@@ -1,12 +1,20 @@
 use rustful::{Server, Handler, Context, Response, TreeRouter, HttpResult};
 use rustful::server::Listening;
+use rustful::server::Global;
+use std::sync::Mutex;
+
 use flow::Blender;
 use query::DataType;
 use std::collections::HashMap;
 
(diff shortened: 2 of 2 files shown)


src/web/main.rs
```diff
@@ -38,9 +43,22 @@ fn main() {
...
         mig.maintain(karma, 0);
+
+        // assign everything to same domain
+        let domain = mig.add_domain();
+        mig.assign_domain(awvc, domain);
+        mig.assign_domain(karma, domain);
+        mig.assign_domain(vc, domain);
+        mig.assign_domain(article, domain);
+        mig.assign_domain(user, domain);
+        mig.assign_domain(vote, domain);
```
jonhoo: Any particular reason you want them all to be in the same domain?

src/web/mod.rs
```diff
@@ -1,14 +1,22 @@
...
+struct GetEndpoint<F> {
     arguments: Vec<String>,
     f: F,
 }
 
+struct PutEndpoint<Mutator> {
+    arguments: Vec<String>,
+    mutator: Mutator,
+}
+
```
jonhoo: Yeah, this is a good change. Should probably have done this a while ago.

src/web/mod.rs
```diff
@@ -50,21 +58,23 @@ pub fn run(soup: Blender) -> HttpResult<Listening> {
     };
 
     for (path, ep) in ins.into_iter() {
-        use std::sync::Mutex;
-        let put = Mutex::new(ep.f);
+        // use std::sync::Mutex;
```
jonhoo: Why would you rather have the import at the top? It's fine, I guess, but then you should remove this line entirely.

src/web/mod.rs
```diff
@@ -50,21 +58,23 @@ pub fn run(soup: Blender) -> HttpResult<Listening> {
...
+
         let args = ep.arguments;
+
+        let put = Mutex::new(Box::new(ep.mutator));
```
jonhoo: Can you leave this line above `let args = ep.arguments` like it was to simplify the diff?

src/web/mod.rs
jonhoo: Ohhh, nvm, you're now also using `Mutex` for `global`. Okay, just delete this line then.

src/web/main.rs
larat7: not really, just for the sake of simplicity

src/web/mod.rs
larat7: because on line 117, we need to Box a Mutex with the Blender to keep it in the global accessible data for the server. without that the blender goes out of scope and channels get closed.

src/web/main.rs
jonhoo: You might just want to leave all `assign_domain` things out entirely then. The nodes will automatically be assigned to separate domains, which is fine for low-throughput applications.

src/web/mod.rs
```diff
@@ -104,6 +111,7 @@ pub fn run(soup: Blender) -> HttpResult<Listening> {
     Server {
             handlers: router,
             host: 8080.into(),
+            global: Global::from(Box::new(Mutex::new(soup))),
```
jonhoo: Actually, why does the soup need to be boxed here? Given that you never *use* `global`, it should be fine to just do `Global::from(soup)`, no?

src/web/main.rs
larat7: Ok, did not know that! I thought we had to manually assign things to domains. Thanks!

src/web/mod.rs
larat7: we tried and the compiler was very sad. 

```error[E0277]: the trait bound `rustful::server::Global: std::convert::From<flow::Blender>` is not satisfied```

jonhoo: Hmm.. The docs say that [it requires](https://doc.rust-lang.org/nightly/core/convert/trait.From.html#tymethod.from) a `Box`, so seems like you'd need `Box::new(soup)`. But the `Mutex` shouldn't be necessary?

jonhoo: Ugh, no, you're right, they also require `Send` and `Sync` :(

jonhoo: I guess it's fine to use a `Mutex` here given that we don't actually ever access it. Just feels dirty...

larat7: I agree, just not sure what would be the cleanest rust approach

jonhoo: Eh, it's fine. Keep it as it is.

PR #8, opened 2017-03-13
Add piazza benchmark

larat7:
Benchmark for Piazza application.

This measures two types of benchmak ```migration``` and ```write```.

```migration``` fills the db with users, classes, posts and relations between users and classes (```taking``` base node). Then, it times how long it takes to login a user. Each user login corresponds to one migration that creates a "parallel universe" for that user.

```write``` prepopulates the db similarly. Then, it logs some users and it measures the time it takes to perform a batch of writes. 

There are two important configurations in this benchmark ```dcfg``` and ```fanout```.  ```dcfg``` specifies the domain configuration of the graph, either a single domain or a domain per user. ```fanout``` specifies the fanout of a write passing through the ```taking``` base node (how many classes each user is taking).

### DIFF
--- a/benchmarks/piazza/piazza.rs
+++ b/benchmarks/piazza/piazza.rs
@@ -228,8 +228,7 @@ fn main() {
         "all" => populate_taking(nclasses, nusers, taking_putter, Fanout::All),
         "few" => populate_taking(nclasses, nusers, taking_putter, Fanout::Few),
         _ => {
-            println!("Invalid fanout configuration");
-            return
+            unreachable!();
         }
     }
 
@@ -238,8 +237,7 @@ fn main() {
         "single"  => domain_config = DomainConfig::Single,
         "peruser" => domain_config = DomainConfig::PerUser,
         _ => {
-            println!("Invalid domain configuration");
-            return
+            unreachable!();
         }
     }
 
@@ -285,8 +283,7 @@ fn main() {
                 app.log_user(uid.into(), &domain_config);
             },
             _ => {
-                println!("wrong benchmark!");
-                return
+                unreachable!();
             }
         };
 
--- a/benchmarks/piazza/piazza.rs
+++ b/benchmarks/piazza/piazza.rs
@@ -243,23 +243,14 @@ fn main() {
         }
     }
 
-    match benchmark.as_ref() {
-        "migration" => {
-            for pid in 0..nposts {
-                post_putter.put(vec![
-                    pid.into(),
-                    (pid % nclasses).into(),
-                    (pid % nusers).into(),
-                    "post".into()
-                    ]);
-            }
-        },
-        "write" => {
-
-        }
-        _ => {
-            println!("wrong benchmark!");
-            return
+    if benchmark == "migration" {
+        for pid in 0..nposts {
+            post_putter.put(vec![
+                pid.into(),
+                (pid % nclasses).into(),
+                (pid % nusers).into(),
+                "post".into()
+                ]);
         }
     }
 
@@ -273,19 +264,17 @@ fn main() {
     }
 
     println!("Starting benchmark...");
-    for uid in 0..(nusers) {
+    for uid in 0..nusers {
         let start;
         let end;
-
+        start = time::Instant::now();
         match benchmark.as_ref() {
             "migration" => {
-                start = time::Instant::now();
                 app.log_user(uid.into(), &domain_config);
 
                 end = time::Instant::now().duration_since(start);
             },
             "write" => {
-                start = time::Instant::now();
                 for i in 0..1000 {
                     post_putter.put(vec![i.into(), (i % nclasses).into(), (i % nusers).into(), "post".into()]);
                 }
--- a/benchmarks/piazza/piazza.rs
+++ b/benchmarks/piazza/piazza.rs
@@ -179,17 +179,14 @@ fn main() {
             for pid in 0..nposts {
                 post_putter.put(vec![
                     pid.into(),
-                    (pid / nclasses).into(),
-                    (pid / nusers).into(),
+                    (pid % nclasses).into(),
+                    (pid % nusers).into(),
                     "post".into()
                     ]);
             }
         },
         "write" => {
-            // login nusers
-            for i in 0..nusers {
-                app.log_user(i.into());
-            }
+
         }
         _ => {
             println!("wrong benchmark!");
@@ -219,12 +216,14 @@ fn main() {
             },
             "write" => {
                 start = time::Instant::now();
-                for cid in 0..nclasses {
-                    post_putter.put(vec![0.into(), cid.into(), uid.into(), "post".into()]);
+                for i in 0..1000 {
+                    post_putter.put(vec![i.into(), (i % nclasses).into(), (i % nusers).into(), "post".into()]);
                 }
                 end = time::Instant::now().duration_since(start);
 
                 thread::sleep(time::Duration::from_millis(1000));
+
+                app.log_user(uid.into());
             },
             _ => {
                 println!("wrong benchmark!");
--- a/benchmarks/piazza/piazza.rs
+++ b/benchmarks/piazza/piazza.rs
@@ -21,6 +21,16 @@ pub struct Piazza {
     domain: Index,
 }
 
+pub enum DomainConfig {
+    Single,
+    PerUser
+}
+
+enum Fanout {
+    All,
+    Few
+}
+
 impl Piazza {
     // Create the base nodes for our Piazza application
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: benchmarks/piazza/piazza.rs [hunk 1]
```diff
@@ -0,0 +1,332 @@
...
+            soup: g,
+            user: user,
+            post: post,
+            class: class,
+            taking: taking,
+            domain: base_domain,
+        }
+    }
+
+    pub fn log_user(&mut self, uid: DataType, domain_config: &str) {
```
jonhoo: You might want to make `domain_config` an `enum`

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+    }
+}
+
+fn populate_classes(nclasses: i64, class_putter: Mutator) {
+    for i in 0..nclasses {
+        class_putter.put(vec![i.into(), i.into()]);
+    }
+}
+
+fn populate_taking(nclasses: i64, nusers: i64, taking_putter: Mutator, fanout: &str) {
```
jonhoo: Same here, `fanout` should be an `enum`

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+
+        }
+        _ => {
+            println!("wrong benchmark!");
+            return
+        }
+    }
+
+    println!("Finished seeding! Sleeping for 1 second...");
+    thread::sleep(time::Duration::from_millis(1000));
```
jonhoo: I think we could probably reduce this to, say, 100ms. Should be plenty of time to drain the queues.

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    println!("Starting benchmark...");
+    for uid in 0..(nusers) {
+        let start;
+        let end;
+
+        match benchmark.as_ref() {
+            "migration" => {
+                start = time::Instant::now();
```
jonhoo: this line can go outside the `match`

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+    thread::sleep(time::Duration::from_millis(1000));
+
+
+    let mut times = Vec::new();
+    if csv {
+       File::create("out.csv").unwrap();
+    }
+
+    println!("Starting benchmark...");
+    for uid in 0..(nusers) {
```
jonhoo: why `0..(nusers)` and not `0..nusers`?

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+    let user_putter = app.soup.get_mutator(app.user);
+    let taking_putter = app.soup.get_mutator(app.taking);
+    let post_putter = app.soup.get_mutator(app.post);
+
+    println!("Seeding...", );
+    populate_users(nusers, user_putter);
+    populate_classes(nclasses, class_putter);
+    populate_taking(nclasses, nusers, taking_putter, fanout);
+
+    match benchmark.as_ref() {
```
jonhoo: Sounds like this could instead just be
```rust
if benchmark.as_ref() == "migration" {
    // ...
}
```
instead

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+
+                thread::sleep(time::Duration::from_millis(1000));
+
+                app.log_user(uid.into(), domain_config);
+            },
+            _ => {
+                println!("wrong benchmark!");
+                return
+            }
+        };
```
jonhoo: If I'm reading this right, couldn't you do:
```rust
let start = time::Instant::now();
if benchmark.as_ref() == "write" {
    for i in 0..1000 {
        post_putter.put(vec![i.into(), (i % nclasses).into(), (i % nusers).into(), "post".into()]);
    }
    thread::sleep(time::Duration::from_millis(100)); // 100ms should be fine
}
let end1 = time::Instant::now().duration_since(start);
let start = time::Instant::now();
app.log_user(uid.into(), domain_config);
let end2 = time::Instant::now().duration_since(start);
let end = match benchmark.as_ref() {
    "write" => end1,
    "migration" => end2,
    _ => unreachable!(),
};
```

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+
+fn max_duration(stats: &Vec<f64>) -> f64 {
+    let max = stats.iter().fold(0f64, |acc, el| {
+        f64::max(acc, *el)
+    });
+
+    max
+}
+
+fn min_duration(stats: &Vec<f64>) -> f64 {
```
jonhoo: Is this not just `stats.iter().min().unwrap()`?

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+    let min_start = stats[0];
+    let min = stats.iter().fold(min_start, |acc, el| {
+        f64::min(acc, *el)
+    });
+
+    min
+}
+
+
+fn avg(stats: &Vec<f64>) -> f64 {
```
jonhoo: == `stats.iter().sum() / stats.len()`?

benchmarks/piazza/piazza.rs
```diff
@@ -0,0 +1,332 @@
...
+    println!("{:?} results ", benchmark);
+    println!("avg: {:?}", avg(&times) );
+    println!("max: {:?}", max_duration(&times) );
+    println!("min: {:?}", min_duration(&times) );
+
+    println!("Done with benchmark.");
+
+}
+
+fn max_duration(stats: &Vec<f64>) -> f64 {
```
jonhoo: == `stats.iter().max().unwrap()`

jonhoo:
Some relatively minor comments inline in the code. Apart from that, I'm pretty happy to merge this soon. @larat7 do you have other outstanding changes that haven't been made part of this PR yet?

larat7:
Nope. I will address the comments and merge this. Also, it's unclear why the tests fail on travis, but not locally.

benchmarks/piazza/piazza.rs
larat7: ```the trait `std::cmp::Ord` is not implemented for `f64` ```

benchmarks/piazza/piazza.rs
larat7: same as ```min```

benchmarks/piazza/piazza.rs
larat7: Sure, but is that more readable than separating out the two benchmarks completely?

jonhoo: It avoids replicating the `app.log_user` call, which is nice to ensure we don't accidentally make them different somehow. The `end1` `end2` game is unfortunate though. I'm okay with either solution.

benchmarks/piazza/piazza.rs
jonhoo: Ohhhh. Ugh. Damn NaN, eh @ms705? Okay, I guess we'll just have to leave it the way it is for now.

benchmarks/piazza/piazza.rs
```diff
@@ -222,7 +224,24 @@ fn main() {
...
+
+    let domain_config;
+    match domain_config_str.as_ref() {
+        "single"  => domain_config = DomainConfig::Single,
+        "peruser" => domain_config = DomainConfig::PerUser,
+        _ => {
+            println!("Invalid domain configuration");
+            return
+        }
+    }
```
jonhoo: Just for reference, you can write this as:
```rust
let domain_config = match domain_config_str.as_ref() {
  "single" => DomainConfig::Single,
  "peruser" => DomainConfig::PerUser,
  _ => { ... }
};
```

benchmarks/piazza/piazza.rs
```diff
@@ -261,7 +280,7 @@ fn main() {
         match benchmark.as_ref() {
             "migration" => {
                 start = time::Instant::now();
-                app.log_user(uid.into(), domain_config);
+                app.log_user(uid.into(), &domain_config);
```
jonhoo: If you `#[derive(Clone, Copy)]` for `DomainConfig`, you could leave off the `&`s here, and then `*` for the corresponding `match`es.

jonhoo:
> it's unclear why the tests fail on travis, but not locally
It's the timeouts we use that are too small for Travis. I'll take a look at increasing them so we avoid these spurious failures. Ignore the failures for now.

PR #9, opened 2017-04-10
Filter reordering

larat7:
Reorder filters above aggregations that aggregate over common columns. This allow us to remove some hotcrap queries from the blacklist.

### DIFF
--- a/src/sql/mir.rs
+++ b/src/sql/mir.rs
@@ -761,6 +761,28 @@ impl SqlToMirConverter {
                 }
             }
 
+
+            // 3. Get columns used by each filter node.
+            let mut filter_columns = HashMap::new();
+            let mut filter_nodes = Vec::new();
+            let mut moved_filters = Vec::new();
+            let mut sorted_rels: Vec<&String> = qg.relations.keys().collect();
+            sorted_rels.sort();
+            for rel in &sorted_rels {
+                let qgn = &qg.relations[*rel];
+                if *rel != "computed_columns" {
+                    if !qgn.predicates.is_empty() {
+                        for cond in &qgn.predicates {
+                          let col = match *cond.left.as_ref() {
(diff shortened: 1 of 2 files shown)


src/sql/mir.rs
```diff
@@ -761,6 +761,28 @@ impl SqlToMirConverter {
...
+
+            // 3. Get columns used by each filter node.
+            let mut filter_columns = HashMap::new();
+            let mut filter_nodes = Vec::new();
+            let mut moved_filters = Vec::new();
+            let mut sorted_rels: Vec<&String> = qg.relations.keys().collect();
+            sorted_rels.sort();
+            for rel in &sorted_rels {
+                let qgn = &qg.relations[*rel];
+                if *rel != "computed_columns" {
```
jonhoo: Can you invert this and use `continue`?

src/sql/mir.rs
```diff
@@ -761,6 +761,28 @@ impl SqlToMirConverter {
...
+                    if !qgn.predicates.is_empty() {
```
jonhoo: I don't think this is necessary? The `for` loop will just do 0 iterations...

larat7: this is the same code that was being used to create filter nodes. the same line comes with this comment https://github.com/mit-pdos/distributary/blob/447d8a8538ef510ecd8ed98c9805df990ac13567/src/sql/mir.rs#L900 not sure if it is still a valid concern

jonhoo: That code looks different from your though. It doesn't just iterate over the predicates.

larat7: The iteration is done inside `make_filter_nodes`

ms705: I think it can be removed in this case (where we're only extracting relevant columns), although not in the case linked (where other datastructures are modified inside the the conditional but outside the loop).

PR #10, opened 2017-04-11
Persistent log

jmftrindade:


jonhoo:
This merge seems wrong in a couple of places. For example, `targets/netsoup.rs` shouldn't exist..?

jmftrindade:
Yup, it looks like these 3 extra files are artifacts of previous merges (it's likely they stayed around because the original changes on master moved them elsewhere, and they weren't being used anymore?):

- benchmarks/vote/targets/netsoup.rs
- src/flow/data.rs
- src/flow/sql_to_flow.rs

Will remove these from the branch, make sure that all tests still pass, and update this request.

jmftrindade:
Alright, pull request updated.

Cargo.toml
```diff
@@ -9,9 +9,9 @@ b_memcached = ["memcached-rs"]
 b_postgresql = ["postgres"]
 b_mysql = ["mysql"]
 b_mssql = ["futures", "futures-state-stream", "tiberius", "tokio-core"]
-b_netsoup = ["futures", "tokio-core", "tarpc", "tarpc-plugins", "serde", "serde_derive"]
+b_netsoup = ["futures", "tokio-core", "tarpc", "tarpc-plugins"]
 b_hybrid = ["mysql", "memcached-rs"]
-default = ["web", "b_netsoup", "b_memcached", "b_mysql", "b_mssql", "b_hybrid"]
+default = ["web", "b_netsoup", "b_memcached", "b_mysql", "b_mssql", "b_hybrid", "serde", "serde_derive"]
```
jonhoo: At this point you can probably just make `serde` and `serde_derive` be non-optional dependencies, and then not list them under `default`.

src/ops/base.rs
```diff
@@ -1,39 +1,221 @@
...
+    /// Synchronous writes: forces every record to be written to disk before it is emitted further
+    /// into the data-flow graph. Strong guarantees (writes are never lost), but high performance
+    /// penalty.
+    SyncImmediately,
+}
+
 impl Base {
     /// Create a base node operator.
-    pub fn new(primary_key: Vec<usize>) -> Self {
+    pub fn new(primary_key: Vec<usize>, durability: BaseDurabilityLevel) -> Self {
```
jonhoo: Can we instead add another method, `new_durable`, that takes two arguments, leaving `new` to just take one, and give no durability? Would also reduce the diff since all the tests don't have to change :p

src/flow/core/processing.rs
```diff
@@ -84,7 +84,7 @@ pub trait Ingredient
                 data: prelude::Records,
                 domain: &prelude::DomainNodes,
                 states: &prelude::StateMap)
-                -> prelude::Records;
+                -> Option<prelude::Records>;
```
jonhoo: Agh, that's going to be annoying to incorporate into `partial-materialization`, but I guess it's necessary here.

src/flow/domain/single.rs
```diff
@@ -161,7 +161,7 @@ impl NodeDescriptor {
             }
             flow::node::Type::Internal(ref mut i) => {
                 let from = m.link().src;
-                m.map_data(|data| i.on_input(from, data, nodes, state));
+                m.map_data(|data| i.on_input(from, data, nodes, state).unwrap());
```
jonhoo: Why is this `unwrap()` okay? Can't a `Base` be processed?

src/flow/mod.rs
```diff
@@ -463,7 +463,7 @@ impl<'a> Migration<'a> {
                 self.mainline.ingredients.add_edge(*parent.as_global(), ni, false);
             }
         }
-        // and tell the caller its id
+
```
jonhoo: Why?

src/flow/payload.rs
```diff
@@ -74,6 +74,9 @@ pub enum Packet {
         parents: Vec<LocalNodeIndex>,
     },
 
+    /// Instruct base nodes to flush any buffered writes.
```
jonhoo: Can we cut this entirely for now?

src/lib.rs
```diff
@@ -175,7 +175,8 @@
 //! # let mut g = Blender::new();
 //! # let article = {
 //! # let mut mig = g.start_migration();
-//! # let article = mig.add_ingredient("article", &["id", "title"], Base::default());
+//! # let article = mig.add_ingredient("article", &["id", "title"],
+//! #                                  Base::default());
```
jonhoo: Why?

src/ops/base.rs
```diff
@@ -1,39 +1,221 @@
...
 /// type corresponding to the node's type.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Base {
+    buffered_writes: Option<Records>,
+    durability: BaseDurabilityLevel,
+    durable_log: Option<BufWriter<File, WhenFull>>,
+    durable_log_path: Option<PathBuf>,
+    global_address: Option<NodeAddress>,
+    last_flushed_at: Option<Instant>,
```
jonhoo: A lot of `Option` here. Could we instead have a `struct BaseBufferState`, and then have `Base` just have one field of type `Option<BaseBufferState>`?

src/ops/base.rs
```diff
@@ -1,39 +1,221 @@
...
+            BaseDurabilityLevel::None => panic!("tried to persist non-durable base node!"),
+            BaseDurabilityLevel::Buffered |
+            BaseDurabilityLevel::SyncImmediately => {
+                self.ensure_log_writer();
+                serde_json::to_writer(&mut self.durable_log.as_mut().unwrap(), &records)
+                    .unwrap();
+                // XXX(malte): we must deconstruct the BufWriter in order to get at the contained
+                // File (on which we can invoke sync_data(), only to then reassemble it
+                // immediately. I suspect this will work best if we flush after accumulating
+                // batches of writes.
```
jonhoo: Does dropping a `BufWriter` not flush the underlying stream? If so, `drop(xxx.take())` should be sufficient.

src/ops/base.rs
```diff
@@ -1,39 +1,221 @@
...
+                Err(reason) => {
+                    panic!("Unable to open durable log file {}, reason: {}",
+                           path.display(), reason)
+                }
+                Ok(file) => file,
+            };
+
+            match self.durability {
+                BaseDurabilityLevel::None => {
+                    self.durable_log = None;
```
jonhoo: In this case it seems like we also shouldn't even create the log file, no?

src/ops/base.rs
```diff
@@ -1,39 +1,221 @@
...
+                }
+            }
+        }
+    }
+
+    /// XXX: This should only be used by tests.  We don't hide it behind cfg test, however, since it
+    /// needs to be available for integration tests, which get compiled against the regular build.
+    pub fn delete_durable_log(&mut self) {
+        // Cleanup any durable log files.
+        if self.durable_log_path.is_some() {
```
jonhoo: `if let Some(ref path) = self.durable_log_path` avoids the `unwrap` below.

src/ops/base.rs
```diff
@@ -67,8 +249,46 @@ impl Ingredient for Base {
                 rs: Records,
                 _: &DomainNodes,
                 state: &StateMap)
-                -> Records {
-        rs.into_iter()
+                -> Option<Records> {
+        // Write incoming records to log before processing them if we are a durable node.
+        let records_to_return;
+        let mut copy_rs = rs.clone();
```
jonhoo: Why is this `clone` necessary? It seems pretty costly, especially because it's not just executed for durable nodes.

src/ops/base.rs
```diff
@@ -67,8 +249,46 @@ impl Ingredient for Base {
...
+                //
+                // 1. Enough time has passed since the last time we flushed.
+                // 2. Our buffer of write records reaches capacity.
+                let num_buffered_writes = self.buffered_writes.as_ref().unwrap().len();
+                let has_reached_capacity = num_buffered_writes + rs.len() >= BUFFERED_WRITES_CAPACITY;
+                let elapsed = self.last_flushed_at.unwrap().elapsed();
+                let has_reached_time_limit = elapsed >= Duration::from_millis(BUFFERED_WRITES_FLUSH_INTERVAL_MS);
+
+                if has_reached_capacity || has_reached_time_limit {
+                    self.buffered_writes.as_mut().unwrap().append(copy_rs.as_mut());
```
jonhoo: Can't this just take `rs`?

src/ops/base.rs
```diff
@@ -67,8 +249,46 @@ impl Ingredient for Base {
...
+
+                    let copy_buffered_writes = self.buffered_writes.as_mut().unwrap().clone();
```
jonhoo: Can't this just `drain(..)`?

src/ops/base.rs
```diff
@@ -67,8 +249,46 @@ impl Ingredient for Base {
...
+                    self.flush();
+
+                    // This returns everything that was buffered, plus the newly inserted records.
+                    records_to_return = Some(copy_buffered_writes);
+                } else {
+                    // Otherwise, buffer the records and don't send them downstream.
+                    self.buffered_writes.as_mut().unwrap().append(copy_rs.as_mut());
```
jonhoo: This can also just consume `rs`

src/ops/base.rs
```diff
@@ -67,8 +249,46 @@ impl Ingredient for Base {
...
+                    return Some(Records::default())
```
jonhoo: Shouldn't this return `None`?

src/ops/base.rs
```diff
@@ -67,8 +249,46 @@ impl Ingredient for Base {
...
+                }
+            },
+            BaseDurabilityLevel::SyncImmediately => {
+                self.persist_to_log(&rs);
+                records_to_return = Some(rs.clone());
```
jonhoo: Why does this need to clone?

src/ops/base.rs
```diff
@@ -67,8 +249,46 @@ impl Ingredient for Base {
...
+            },
+            BaseDurabilityLevel::None => {
+                records_to_return = Some(rs.clone());
```
jonhoo: ditto

src/ops/mod.rs
```diff
@@ -190,7 +190,9 @@ pub mod test {
                 .clear();
         }
 
-        pub fn one<U: Into<Records>>(&mut self, src: NodeAddress, u: U, remember: bool) -> Records {
+        pub fn one<U: Into<Records>>(&mut self,
```
jonhoo: Unrelated?

src/ops/mod.rs
```diff
@@ -204,7 +206,8 @@ pub mod test {
                 .as_local()]
                     .borrow_mut()
                     .inner
-                    .on_input(src, u.into(), &self.nodes, &self.states);
+                    .on_input(src, u.into(), &self.nodes, &self.states)
```
jonhoo: Unrelated?

src/ops/mod.rs
```diff
@@ -190,7 +190,9 @@ pub mod test {
                 .clear();
         }
 
-        pub fn one<U: Into<Records>>(&mut self, src: NodeAddress, u: U, remember: bool) -> Records {
+        pub fn one<U: Into<Records>>(&mut self,
+                                     src: NodeAddress,
+                                     u: U, remember: bool) -> Records {
```
jonhoo: Just formatting?

tests/lib.rs
```diff
@@ -71,6 +73,178 @@ fn it_works() {
...
 #[test]
+fn it_propagates_writes_w_durability_sync_immediately() {
+    use distributary::{Base, BaseDurabilityLevel};
+
+    // set up graph
+    let mut g = distributary::Blender::new();
+    let (a, b, cq) = {
+        let mut mig = g.start_migration();
+        let a = mig.add_ingredient("a", &["a", "b"],
+                                   Base::new(vec![0], BaseDurabilityLevel::SyncImmediately).delete_log_on_drop());
```
jonhoo: To make this look a bit neater, you may want to assign `BaseDurabilityLevel::SyncImmediately` to a variable with a short name instead :)

tests/lib.rs
```diff
@@ -136,16 +310,16 @@ fn shared_interdomain_ancestor() {
     // send a value on a
     muta.put(vec![id.clone(), 2.into()]);
     assert_eq!(bq.recv_timeout(time::Duration::from_millis(100)),
-               Ok(vec![vec![id.clone(), 2.into()].into()]));
+    Ok(vec![vec![id.clone(), 2.into()].into()]));
```
jonhoo: just formatting?

tests/lib.rs
```diff
@@ -208,12 +382,15 @@ fn it_works_w_mat() {
...
+
     // set up graph
     let mut g = distributary::Blender::new();
     let (a, b, cq) = {
         let mut mig = g.start_migration();
-        let a = mig.add_ingredient("a", &["x", "y"], distributary::Base::new(vec![1]));
-        let b = mig.add_ingredient("b", &["_", "x", "y"], distributary::Base::new(vec![2]));
+        let a = mig.add_ingredient("a", &["x", "y"], Base::new(vec![1], BaseDurabilityLevel::None));
+        let b = mig.add_ingredient("b", &["_", "x", "y"],
+                                   Base::new(vec![2], BaseDurabilityLevel::None));
```
jonhoo: With the `::new` change, this diff will disappear.

tests/lib.rs
```diff
@@ -241,7 +418,7 @@ fn it_works_deletion() {
     use distributary::StreamUpdate::*;
     muta.delete(vec![2.into()]);
     assert_eq!(cq.recv(),
-               Ok(vec![DeleteRow(Arc::new(vec![1.into(), 2.into()]))]));
+    Ok(vec![DeleteRow(Arc::new(vec![1.into(), 2.into()]))]));
```
jonhoo: Just formatting?

tests/lib.rs
```diff
@@ -623,7 +800,7 @@ fn transactional_migration() {
 
     // check that c got them
     assert_eq!(cq(&3.into()).unwrap().0,
-               vec![vec![3.into(), 5.into()], vec![3.into(), 6.into()]]);
+    vec![vec![3.into(), 5.into()], vec![3.into(), 6.into()]]);
```
jonhoo: Just formatting?

tests/lib.rs
```diff
@@ -994,7 +1171,7 @@ fn live_writes() {
         let mut mig = g.start_migration();
 
         // add vote base table
-        vote = mig.add_ingredient("vote", &["user", "id"], distributary::Base::default());
+        vote = mig.add_ingredient("vote", &["user", "id"], Base::default());
```
jonhoo: I'd prefer it if you removed this diff, just to make the PR smaller.

jonhoo:
Okay, @jmftrindade, I think that's all I wanted to say :p Overall looks pretty good — this is mostly just nit.

Cargo.toml
jmftrindade: Done

src/ops/base.rs
jmftrindade: Done

src/flow/mod.rs
jmftrindade: Reverted

src/lib.rs
jmftrindade: Reverted

tests/lib.rs
jmftrindade: Reverted

tests/lib.rs
jmftrindade: Reverted

tests/lib.rs
jmftrindade: Done

tests/lib.rs
jmftrindade: Done

src/ops/mod.rs
jmftrindade: This one removes the semicolon from this line, and adds an unwrap() call right after.

src/ops/mod.rs
jmftrindade: Reverted

src/ops/base.rs
```diff
@@ -1,39 +1,234 @@
...
+    /// penalty.
+    SyncImmediately,
+}
+
 impl Base {
-    /// Create a base node operator.
+    /// Create a non-durable base node operator.
     pub fn new(primary_key: Vec<usize>) -> Self {
         Base {
+            buffered_writes: Some(Records::default()),
```
jonhoo: Personal preference, but I'd prefer if this and `new_durable` both internally called a single method that takes an `Option<BaseDurabilityLevel>`. That way we don't have to repeat the initialization of all the fields.

src/ops/base.rs
```diff
@@ -1,39 +1,234 @@
...
+            last_flushed_at: Some(Instant::now()),
+            primary_key: Some(primary_key),
+            should_delete_log_on_drop: false,
+            unique_id: ProcessUniqueId::new(),
+            us: None,
+        }
+    }
+
+    /// Create durable base node operator.
+    pub fn new_durable(primary_key: Vec<usize>, durability: BaseDurabilityLevel) -> Self {
```
jonhoo: Couldn't we do the `ensure_log_writer` stuff here? And that way remove the call to it entirely from `on_input`?

src/ops/base.rs
jonhoo: Or alternatively, they could both construct using `Base::default()`, and then modify the appropriate fields.

src/flow/domain/single.rs
jmftrindade: Base::on_input is returning Option<Records> now, hence the unwrap().  Should we have a match here instead?

jonhoo: Well, `unwrap` indicates that you believe `on_input` can *never* return `None` in this context. Why is that?

src/ops/base.rs
jmftrindade: Do you think it will still be better than what we have now, given that not all durable fields are always consistently None / Some?  For example, what we currently have in Base::new_durable for the durability state fields is:
...
            buffered_writes: Some(Records::default()),
            durability: Some(durability),
            durable_log: None,
            durable_log_path: None,
            last_flushed_at: Some(Instant::now()),
...

Which means we can't make the fields required on BaseDurabilityState, i.e.,:

struct BaseDurabilityState {
    buffered_writes: Records,
    durability: BaseDurabilityLevel,
    durable_log: BufWriter<File, WhenFull>,
    durable_log_path: PathBuf,
    last_flushed_at: Instant,
}

So we'd have 2 layers of optional fields.  Wouldn't that be weird?

jonhoo: Hmm.. Isn't `buffered_writes` and `durability` only interesting if `BaseDurability != None`, and therefore it's fine for them to be inside `BaseDurabilityState` without an `Option`? Same for `last_flushed_at`, which is only used if durability `!= None`?

src/ops/base.rs
jmftrindade: The durable log path filename relies on self.us local id, which is not available at construction time: it is only provided by the time we get to Base::on_commit.

jonhoo: Ah, that's a shame. But fair enough, I guess.

src/ops/base.rs
jmftrindade: I see.  I think `durable_log_path` would still need to be optional, since at construction time we don't have the Base node's "us" id (it's only provided at `::on_commit`).  Because `durable_log_path` is not known at creation time, `durable_log` would also need to be optional -- otherwise if we instead relied on some default value for path, we'd unnecessarily create a durable log file that's not gonna be used.

jonhoo: Oh, so this stems from the same issue below of not knowing the node's ID until later. That's unfortunate. I think I'd still prefer `Option`s inside of a `BaseDurabilityState`, but you're right that it's less elegant. I wonder if there's a way we could name `Base` nodes when they're created instead of relying on `self.us`. That would mean we could instantiate all of this at creation time, get rid of the `Option`s, and also get rid of the call to `ensure_log`...

src/ops/base.rs
jmftrindade: Yeah, this code looks weird.  So the caller (`::persist_to_log`) was already taking care of calling `::ensure_log_writer` only for the appropriate durability levels.  Moved `::ensure_log_writer`'s match one level up, and made the `None` option unreachable.

src/ops/base.rs
jmftrindade: Yes, naming `Base` nodes at creation time would be ideal, and it would also make it easier for a recovery manager to restore `Base` nodes -- since now an external entity can keep track of the global_id -> Base node mapping, it knows where to fetch the corresponding durable log for a given node.

How does global id generation currently work right now?  Could we expose that as a thread-safe singleton somewhere, so that every time we call `Base::new` it gets assigned an unique pseudo random number (LCG, xorshift, etc)?

jonhoo: Currently each node is assigned a `NodeIndex` when it is inserted into the `petgraph::Graph`, so it's monotonically increasing at `add_ingredient` time. An alternative would be to add a required `name` argument to `Base::new`, which we require to be unique...

src/flow/domain/single.rs
jmftrindade: Because `::on_input` never actually returned None, only Some(empty_records_vector) when buffered writes weren't propagated downstream.  Updated.

src/flow/payload.rs
jmftrindade: Done.

src/ops/base.rs
jmftrindade: But the `NodeIndex` is "local", right?  Adding a `name` argument to `Base::new` that we "assume" to be unique  is a good short-term solution vs perhaps a long-term one being some thread-safe unique PRNG.

The latter makes sure Soup users are not responsible for providing unique names, and that base nodes don't have to check whether their provided `name` is unique.

Alright, will keep this as is.  In a separate pull request -- where we name `Base` nodes -- I can revisit the `Option`s "overload" here :)

src/ops/base.rs
jmftrindade: Done

src/ops/base.rs
jmftrindade: Done

src/ops/base.rs
jmftrindade: `Vec::append` expects a mutable reference, and what we get in the arg list is a vanilla `rs: Records`. If I instead pass rs as a `&mut` (`self.buffered_writes.as_mut().unwrap().append(&mut rs)`), I get a  "cannot borrow mutably" compilation error.

Once I changed `::on_input` to expect a mutable rs instead, then it all worked without the cloning.

src/ops/base.rs
jmftrindade: `drain(..)` would consume the underlying vector, right?  We still need the original buffered records to remain there for the `flush` call on the next line.

src/ops/base.rs
jmftrindade: Done

src/ops/base.rs
jmftrindade: Yeah, that makes a lot more sense than Some(empty Vec).  Otherwise, `::on_input` could just return a regular records Vec.

So I updated callers to explicitly handle `Base::on_input` returning None in the case of buffered writes that are not propagated downstream.

Oddly enough I now get a panic from `Project::resolve_col` via `it_fails_to_resolve_literal`, but this seems unrelated since there's no explicit handling of `Base::on_input` results there?  Why would the check for `self.emit.is_some() && col >= self.emit.as_ref().unwrap().len()` be true if `Base::on_input` can now return `None`?

src/ops/base.rs
jmftrindade: Removed

src/ops/base.rs
jmftrindade: Done

src/ops/mod.rs
jmftrindade: Updated this to explicitly handle `Base::on_input` returning `None` -- instead of implicitly expecting it to return Some(empty vector).

tests/lib.rs
jmftrindade: Done here and for `BaseDurabilityLevel::Buffered` test below.

tests/lib.rs
jmftrindade: Reverted

src/ops/base.rs
jmftrindade: It does, but we still need the original `file` to create a the new `BufWriter`.  So we can instead drop `file.sync_data()` -- since we need the `file` anyway on the next line.

src/ops/base.rs
jmftrindade: Yeah, it isn't.  Wrote it like that because (1) I clearly haven't gotten a good grasp of Rust's borrowing semantics yet, and (2) other permutations of this that I tried all showed cryptic compilation error messages :)  Removed excessive cloning here.

src/flow/domain/single.rs
jonhoo: Hmm, if never returns `None`, why does it need to return an `Option`?

src/ops/base.rs
jonhoo: `NodeIndex` is global for a given execution, `NodeAddress` can be (thread)local or global for the execution. Yeah, I like the idea of moving to `name` — will probably make some of this a bit nicer to work with :) In a separate PR seems fine though.

src/ops/base.rs
jonhoo: Hehe, we've all been there :)

src/ops/base.rs
jonhoo: Ah, no, `drain(..)` crucially does *not* consume the vector, though it does *empty* it. I guess the way you might get away without this clone is to have `self.flush` return the vec it flushed, so that we can then return it again. Sorry to keep poking at these `clones`, but I think it might end up having a decent impact given how all writes go through `Base` nodes.

src/ops/base.rs
jonhoo: Oh, that's weird -- I don't see how that test should be at all related to changes to `Base` :/

jonhoo:
One question: if we *always* end up treating `None` from `on_input` as `Records::default()`, why don't we just keep it a `Records` like it used to be, and return that directly? Will make it easier to merge now following the merge of `partial-materialization` as well..

jonhoo:
Also, don't worry too much about the conflicts with `master`. I can do that merge when it comes to that :)

jonhoo:
Oh, nice, that significantly reduced the number of conflicts too! :D

src/flow/domain/single.rs
jmftrindade: Resolved

src/ops/base.rs
jmftrindade: Ok, so that was actually a false alarm.  That test is not failing, as it expects a panic -- should have been obvious from the `it_fails_to` prefix hehe.  It's just that I run my tests with RUST_BACKTRACE enabled, and since I was getting a backtrace for it, I incorrectly assumed it was a test failure :)

src/ops/base.rs
jmftrindade: Actually done now.

src/ops/base.rs
jmftrindade: Done! 💃

jonhoo:
@jmftrindade I'll take a stab at merging `master` into this at some point tonight, so if you suddenly get a push error, that'll be why. Hopefully it shouldn't be too tricky to rebase any changes you make in the interim on top of that merge.

jonhoo:
Travis failures are due to https://github.com/rust-lang-nursery/rustup.rs/issues/1062

jonhoo:
:tada:

PR #11, opened 2017-04-11
Explicitly designate which base nodes are transactional

fintelia:
If this design looks good, I can update binaries to use the new interface


jonhoo:
Why is `transactional` a property of `Node`, and not of `Base`?

jonhoo:
LGTM. r+

jonhoo:
Merged without merge commit.

PR #12, opened 2017-04-11
combines filter chains into a single node

larat7:
**THIS**
image

**INSTEAD OF** 

image

### DIFF
--- a/src/sql/mir.rs
+++ b/src/sql/mir.rs
@@ -60,13 +60,15 @@ impl SqlToMirConverter {
         SqlToMirConverter { log: log, ..Default::default() }
     }
 
-    /// Converts a condition tree stored in the `ConditionExpr` returned by the SQL parser into a
-    /// vector of conditions that `shortcut` understands.
-    fn to_conditions(&self,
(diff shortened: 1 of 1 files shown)


src/sql/mir.rs
```diff
@@ -60,13 +60,15 @@ impl SqlToMirConverter {
         SqlToMirConverter { log: log, ..Default::default() }
     }
 
-    /// Converts a condition tree stored in the `ConditionExpr` returned by the SQL parser into a
-    /// vector of conditions that `shortcut` understands.
-    fn to_conditions(&self,
+    /// Converts a condition tree stored in the `ConditionExpr` returned by the SQL parser
+    /// and adds its to a vector of conditions that `shortcut` understands.
```
jonhoo: Just btw, we don't use `shortcut` any more :p

jonhoo:
LGTM, but r? @ms705

src/sql/mir.rs
ms705: That's on me :) I wrote that comment and didn't update it.

ms705:
Ok -- I think there are some possible improvements, but it's correct and in the interest of expedience and other merges, we should merge this. Thanks @larat7!

PR #13, opened 2017-04-12
Transactional replay

jonhoo:


src/checktable/mod.rs
```diff
@@ -139,6 +140,10 @@ pub struct CheckTable {
     // For each domain, stores the set of base nodes that it receives updates from.
     domain_dependencies: HashMap<domain::Index, Vec<NodeIndex>>,
 
+    // Domains impacted by each replay path.
+    replay_paths: HashMap<ReplayPath, Vec<domain::Index>>,
```
jonhoo: Down the line, we want to use `VecMap` for this, `domain_dependencies`, `last_replay`, etc.

src/flow/domain/mod.rs
```diff
@@ -349,10 +349,17 @@ impl Domain {
             };
 
             let addr = n.borrow().addr();
-            let mut m = Packet::Transaction {
-                link: Link::new(addr, addr), // TODO: message should be from actual parent, not self.
-                data: data,
-                state: ts.clone(),
+            let mut m = if n.borrow().inner.is_transactional() {
```
jonhoo: I'd like a comment here explaining why this downgrade is okay/necessary.

src/flow/domain/mod.rs
```diff
@@ -593,8 +607,19 @@ impl Domain {
         }
     }
 
-    fn seed_replay(&mut self, tag: Tag, key: &[DataType]) {
-        let (m, source) = match self.replay_paths[&tag] {
+    fn seed_replay(&mut self, tag: Tag, key: &[DataType], transaction_state: Option<ReplayTransactionState>) {
+        if transaction_state.is_none() {
```
jonhoo: I don't understand why this is necessary?

src/flow/domain/mod.rs
```diff
@@ -612,33 +637,44 @@ impl Domain {
...
-                         })
+                             transaction_state: transaction_state,
+                    });
+                    (m, source, false)
+                } else if transaction_state.is_some() {
+                    let m = Some(Packet::ReplayPiece {
+                        link: Link::new(source, path[0].0),
+                        tag: tag,
+                        context: ReplayPieceContext::Partial { for_key: Vec::from(key), ignore: true },
+                        data: Vec::<Record>::new().into(),
```
jonhoo: You can use `Records::default()` here

src/flow/domain/mod.rs
```diff
@@ -612,33 +637,44 @@ impl Domain {
...
                              tag: tag,
-                             context: ReplayPieceContext::Partial { for_key: Vec::from(key) },
+                             context: ReplayPieceContext::Partial { for_key: Vec::from(key), ignore: false },
                              data: Records::from_iter(rs.into_iter().cloned()),
-                         })
+                             transaction_state: transaction_state,
+                    });
+                    (m, source, false)
+                } else if transaction_state.is_some() {
+                    let m = Some(Packet::ReplayPiece {
```
jonhoo: I'd like a brief comment here saying "we need to forward a `ReplayPiece` for the timestamp we claimed"

src/flow/domain/mod.rs
```diff
@@ -612,33 +637,44 @@ impl Domain {
...
         };
 
-        if let Some(m) = m {
+        if is_miss {
+            // we have missed in our lookup, so we have a partial replay through a partial replay
+            // trigger a replay to source node, and enqueue this request.
+            self.on_replay_miss(*source.as_local(), Vec::from(key), tag);
+
+            if let Some(m) = m {
+                self.handle_replay(m);
```
jonhoo: I think our future selves will want a `trace!` here.

jonhoo: Also, can we instead move this out of the `is_miss` conditional, and get rid of the `else` clause?

src/flow/domain/mod.rs
```diff
@@ -829,28 +868,41 @@ impl Domain {
                         }).unwrap();
                     }
                 }
+                mut m @ Packet::ReplayPiece {
+                    context: ReplayPieceContext::Partial { ignore: true, .. }, ..
+                } => {
+                    let mut n = self.nodes[&path.last().unwrap().0.as_local()].borrow_mut();
+                    if n.is_egress() && n.is_transactional() {
+                        // No need to set link src/dst since the egress node will not use them.
```
jonhoo: "We need to propagate this replay even though it contains no data, so that downstream domains don't wait for its timestamp"

src/flow/domain/mod.rs
```diff
@@ -867,6 +919,14 @@ impl Domain {
                             false
                         };
 
+                        if !n.is_transactional() {
+                            if let Packet::ReplayPiece { ref mut transaction_state, .. } = m {
+                                transaction_state.take();
```
jonhoo: "Transactional replays that cross into non-transactional subgraphs should stop being transactional. This is necessary to ensure that they don't end up being buffered, and thus re-ordered relative to subsequent writes to the same key."

jonhoo: Also, shouldn't this modify `is_transactional`?

src/flow/domain/mod.rs
```diff
@@ -937,6 +997,26 @@ impl Domain {
                         drop(n);
 
                         if let Packet::None = m {
+                            if !target && partial_key.is_some() && is_transactional {
```
jonhoo: "The partial replay was captured, but we still need to propagate an (ignored) `ReplayPiece` so that downstream domains don't end up waiting forever for the timestamp we claimed."

src/flow/domain/single.rs
```diff
@@ -224,8 +224,11 @@ impl NodeDescriptor {
                 let from = m.link().src;
 
                 let replay = if let Packet::ReplayPiece {
-                           context: flow::payload::ReplayPieceContext::Partial { ref for_key }, ..
-                       } = *m {
+                    context: flow::payload::ReplayPieceContext::Partial {
+                        ref for_key, ignore
```
jonhoo: Actually, the more I think about it, the more I think `ignore` should be replaced with a `ReplayPieceContext` variant like `TransactionalIgnore`.

src/flow/domain/mod.rs
```diff
@@ -987,7 +1067,8 @@ impl Domain {
                         ReplayPieceContext::Regular { .. } => {
                             debug!(self.log, "batch processed");
                         }
-                        ReplayPieceContext::Partial { for_key } => {
+                        ReplayPieceContext::Partial { for_key, ignore } => {
+                            assert!(!ignore);
```
jonhoo: Actually, the more I think about it, the more I think `ignore` should be replaced with a `ReplayPieceContext` variant like `TransactionalIgnore`.

src/flow/mod.rs
```diff
@@ -22,7 +22,7 @@ pub mod payload;
 pub mod statistics;
 pub mod keys;
 pub mod core;
-mod migrate;
+pub mod migrate;
```
jonhoo: Umm... Why was this necessary?

src/flow/transactions/mod.rs
```diff
@@ -186,6 +204,10 @@ impl DomainState {
                         Packet::CompleteMigration { ingress_from_base, .. } => {
                             Bundle::MigrationEnd(ingress_from_base)
                         }
+                        Packet::ReplayPiece { .. } => {
+                            println!("[{:?} @ {}] next_transaction ← ReplayPiece({})", self.domain_index, self.ts, ts);
```
jonhoo: I assume these shouldn't actually be here?

src/flow/transactions/mod.rs
```diff
@@ -288,7 +319,52 @@ impl DomainState {
...
+                println!("[{:?} @ {}] Replay({})", self.domain_index, self.ts, self.ts);
+                Event::Replay(packet)
+            }
+            Bundle::SeedReplay(tag, key, rts) => {
+                self.ts += 1;
+                self.update_next_transaction();
+                println!("[{:?} @ {}] SeedReplay({})", self.domain_index, self.ts, self.ts);
+                Event::SeedReplay(tag, key, rts)
+            }
             Bundle::Empty => Event::None,
```
jonhoo: Couldn't you have this `return`, and then move all the
```rust
self.ts += 1;
self.update_next_transaction();
```
into a single one at the bottom?

src/flow/domain/mod.rs
jonhoo: Also, this seems to be the only place where `schedule_replay` is called? Shouldn't it also be the case for when you are a `TriggerEndpoint::Start` (i.e., when you're going to send a replay to someone else). Those presumably also need timestamps?

jonhoo: After discussion, this *should* be `TriggerEndpoint::Start`, and *not* `::Local`.

jonhoo:
Also, `Mutator` should disallow transactional writes to non-transactional bases, and vice-versa.

PR #14, opened 2017-04-17
SQL code can add transactional bases nodes

fintelia:
Comments on the API?

### DIFF
--- a/src/sql/mod.rs
+++ b/src/sql/mod.rs
@@ -80,8 +80,8 @@ impl SqlIncorporator {
     }
 
     /// Make any future base nodes added be transactional.
-    pub fn make_transactional(&mut self) {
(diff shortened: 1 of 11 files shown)


src/sql/mod.rs
```diff
@@ -77,6 +79,11 @@ impl SqlIncorporator {
         }
     }
 
+    /// Make any future base nodes added be transactional.
+    pub fn make_transactional(&mut self) {
+        self.transactional = true;
+    }
```
ms705: This presumably means that all future base nodes added will be transactional, whether `recipe::activate()` is called with the `transactional` flag set or not. Is there a reason why we need to make this a flag that can only be flipped once?

fintelia: Fixed.

PR #19, opened 2017-07-30
Correctly handles `and` and `or` predicates

larat7:
Previously, we incorrectly handled `or` predicates and only allowed for one level of nested predicates. 

Now, we support multiple levels of nesting and `and` predicates are constructed by chaining the lefts-side and the right-side nodes, while `or` predicates are constructed by unioning the lefts-side and the right-side nodes.

There are limitations to the way we currently handle `or` predicates, more specifically we **don't** support:
-  `or` expressions with predicates with placeholder parameters, because these expressions are meaningless in the Soup context.
- `or` expressions with join predicates.
- `or` expressions between different tables (e.g ```table1.x = 1 OR table2.y= 42```).

### DIFF
--- a/src/sql/mir.rs
+++ b/src/sql/mir.rs
@@ -455,7 +455,7 @@ impl SqlToMirConverter {
     }
 
     fn make_union_node(
-        &mut self,
+        &self,
         name: &str,
         ancestors: Vec<MirNodeRef>
     ) -> MirNodeRef {
@@ -490,7 +490,7 @@ impl SqlToMirConverter {
     }
 
     fn make_filter_nodes(
-        &mut self,
+        &self,
         name: &str,
         parent: MirNodeRef,
         predicates: &Vec<ConditionTree>,
@@ -821,40 +821,81 @@ impl SqlToMirConverter {
     }
 
     fn make_predicate_nodes(
-        &self, 
+        &self,
+        name: &str,
+        parent: MirNodeRef,
         ce: &ConditionExpression,
+        nc: usize,
     ) -> Vec<MirNodeRef> {
         use nom_sql::ConditionExpression::*;
-        
+
+        let mut pred_nodes: Vec<MirNodeRef> = Vec::new();
         match *ce {
-            ComparisonOp(ct) => {
-                let left = self.make_predicate_nodes(ct.left);
-                let right = self.make_predicate_nodes(ct.right);
-
+            ComparisonOp(ref ct) => {
+                let (left, right);
                 match ct.operator {
-                    And => {
-                        assert!(left.len() > 0, "expected nodes for predicate {:?}", ct.left);
-                        assert!(right.len() > 0, "expected nodes for predicate {:?}", ct.right);
+                    Operator::And => {
+                        left = self.make_predicate_nodes(
+                            name,
+                            parent.clone(),
+                            &*ct.left,
+                            nc
+                        );
 
-                        let last_left = left.last().unwrap();
-                        let first_right = right.first().unwrap();
+                        right = self.make_predicate_nodes(
+                            name,
+                            left.last().unwrap().clone(),
+                            &*ct.right,
+                            nc + left.len()
+                        );
 
-                        first_right.borrow_mut().add_ancestor(last_left);
-                        last_left.borrow_mut().add_child(first_right);
+                        pred_nodes.extend(left.clone());
+                        pred_nodes.extend(right.clone());
 
                     },
-                    Or => {
-                        // join by union node
+                    Operator::Or => {
+                        left = self.make_predicate_nodes(
+                            name,
+                            parent.clone(),
+                            &*ct.left,
+                            nc
+                        );
+
+                        right = self.make_predicate_nodes(
+                            name,
+                            parent.clone(),
+                            &*ct.right,
+                            nc + left.len()
+                        );
+
+                        pred_nodes.extend(left.clone());
+                        pred_nodes.extend(right.clone());
+
+                        let last_left = left.last().unwrap().clone();
+                        let last_right = right.last().unwrap().clone();
+                        let union = self.make_union_node(
+                            &format!("{}_p{}", name, nc + left.len() + right.len()),
+                            vec![last_left, last_right]
+                        );
+
+                        pred_nodes.push(union);
                     }
+                    _ => unreachable!()
                 }
-            }, 
-            LogicalOp(ct) => {
+            },
+            LogicalOp(ref ct) => {
                 // currently, we only support filter logical operations
-                self.make_filter_nodes(ct);
-            }, 
-            NegationOp(ct) => unreachable!("negation should have been removed earlier"), 
-            Base(cb) => unreachable!(""), 
+                pred_nodes = self.make_filter_nodes(
+                    &format!("{}_p{}", name, nc),
+                    parent,
+                    &vec![ct.clone()]
+                );
+            },
+            NegationOp(_) => unreachable!("negation should have been removed earlier"),
+            Base(_) => unreachable!("dangling base predicate"),
         }
+
+        pred_nodes
     }
 
     /// Returns list of nodes added
--- a/src/sql/mir.rs
+++ b/src/sql/mir.rs
@@ -1299,7 +1299,7 @@ impl SqlToMirConverter {
                         };
 
                         let fns = self.make_predicate_nodes(
-                            &format!("q_{:x}_p", qg.signature().hash),
+                            &format!("q_{:x}_n{}_p", qg.signature().hash, new_node_count),
                             parent,
                             p,
                             predicate_nodes.len(),
--- a/src/sql/mir.rs
+++ b/src/sql/mir.rs
@@ -808,6 +808,43 @@ impl SqlToMirConverter {
         )
     }
 
+    fn make_predicate_nodes(
+        &self, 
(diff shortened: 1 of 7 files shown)


larat7:
seems like I broke some other stuff with the Finkelstein algorithm. will look into it tomorrow.

larat7:
@ms705 this should be good now. would you mind taking a look?

src/sql/mir.rs
```diff
@@ -925,33 +1084,32 @@ impl SqlToMirConverter {
             }
 
 
-            // 2. Get columns used by each filter node.
-            let mut filter_columns = HashMap::new();
-            let mut filter_nodes = Vec::new();
-            let mut moved_filters = Vec::new();
+            // 2. Get columns used by each predicate. This will be used to check
+            // if we need to reorder predicates before group_by nodes.
+            let mut pred_cols = HashMap::new();
```
ms705: Does this really contain columns? I read the code below as putting `ConditionExpression` objects from `qgn.predicates` into it.

Edit: Ah, it's a `HashMap<Column, Vec<ConditionExpression>>`, so it stores all predicates with each column contained in the predicates (an inverse mapping). Is it easy to avoid cloning all the predicates every time and to store a `Vec<&ConditionExpression>` instead?

src/sql/mir.rs
```diff
@@ -964,6 +1122,34 @@ impl SqlToMirConverter {
...
+                                    &format!(
+                                        "q_{:x}_n{}",
+                                        qg.signature().hash,
+                                        new_node_count
+                                    ),
+                                    &pred_cols,
+                                    over_col.clone(),
+                                    parent,
+                                    &mut moved_predicates);
+
```
ms705: If I read the code correctly, `reorder_predicates` ends up generating nodes for *each* predicate associated with the `over_col` in `pred_cols`. Won't this end up generating nodes for the same predicate several times over? (since a predicate may be associated with multiple columns in `pred_cols`)

src/sql/query_graph.rs
```diff
@@ -219,39 +225,113 @@ impl QueryGraph {
...
+                &mut new_join,
                 &mut global,
-                &mut params,
+                &mut new_params,
             );
+
+            match ct.operator {
+                Operator::And => {
+                    for (t, ces) in new_local {
+                        assert!(ces.len() <= 2, "can only combine less than 2 ConditionExpression's");
```
ms705: nit: "two or fewer" (it's not strictly less)

src/sql/reuse.rs
```diff
@@ -94,6 +95,54 @@ where
...
+/// Direct elimination for complex predicates with nested `and` and `or` expressions
+fn complex_predicate_implies(np: &ConditionExpression, ep: &ConditionExpression) -> bool {
+    match *ep {
+        LogicalOp(ref ect) => {
+            match *np {
+                LogicalOp(ref nct) => {
+                    if nct.operator == ect.operator {
+                        return (complex_predicate_implies(&*nct.left, &*ect.left) && complex_predicate_implies(&*nct.right, &*ect.right)) ||
+                                (complex_predicate_implies(&*nct.left, &*ect.right) && complex_predicate_implies(&*nct.right, &*ect.left));
+                    }
```
ms705: Nice, this captures the case when the predicates in the `OR` condition are flipped :)

src/sql/reuse.rs
```diff
@@ -320,4 +362,173 @@ mod tests {
...
+            right: Box::new(pd),
+            operator: Operator::Or
+        });
+
+        // a < 80 or a > 20
+        let cp3 = LogicalOp(ConditionTree {
+            left: Box::new(pb),
+            right: Box::new(pa),
+            operator: Operator::Or
+        });
```
ms705: Is it deliberate that this predicate is tautological? (There is no `a` that is greater than 80 and less than 20.) The implications are correct, I was just curious if this is a deliberate choice.

src/sql/reuse.rs
```diff
@@ -320,4 +362,173 @@ mod tests {
...
+            left: Box::new(Base(Field(Column::from("a")))),
+            right: Box::new(Base(Literal(Literal::Integer(80.into())))),
+        });
+
+        // a > 20 and a < 60
+        let cp1 = LogicalOp(ConditionTree {
+            left: Box::new(pa.clone()),
+            right: Box::new(pb.clone()),
+            operator: Operator::And
+        });
```
ms705: See above -- deliberate that this predicate is always `false`?

src/sql/reuse.rs
```diff
@@ -320,4 +362,173 @@ mod tests {
...
+            right: Box::new(pd),
+            operator: Operator::And
+        });
+
+        // a < 60 and a > 20
+        let cp3 = LogicalOp(ConditionTree {
+            left: Box::new(pb),
+            right: Box::new(pa),
+            operator: Operator::And
+        });
```
ms705: See above -- deliberate that this predicate is always `false`?

src/sql/reuse.rs
```diff
@@ -320,4 +362,173 @@ mod tests {
...
+
+
+        assert!(complex_predicate_implies(&pa, &cp1));
+        assert!(complex_predicate_implies(&pb, &cp1));
+        assert!(!complex_predicate_implies(&cp1, &pa));
+        assert!(!complex_predicate_implies(&cp1, &pb));
+    }
+
+    #[test]
+    fn complex_predicate_implication_subset_and() {
```
ms705: Awesome, I love that we cover subset and superset cases too :+1:

src/sql/mir.rs
```diff
@@ -785,6 +808,142 @@ impl SqlToMirConverter {
         )
     }
 
+    fn make_predicate_nodes(
+        &self,
+        name: &str,
+        parent: MirNodeRef,
+        ce: &ConditionExpression,
+        nc: usize,
```
larat7: Yeah, it's supposed to be the count of nodes inside the predicate. We had something similar for filters, so I kept it. I don't think there's any reason why this couldn't be the total node count instead.

src/sql/mir.rs
```diff
@@ -785,6 +808,142 @@ impl SqlToMirConverter {
...
+                },
+                NegationOp(_) => unreachable!("negations should have been eliminated"),
+                _ => (),
+            }
+        }
+
+        cols
+    }
+
+    fn reorder_predicates(
```
larat7: Ah, you're right that the naming here is a bit misleading. We first identify which filters will conflict with group_by nodes and we just created them before the group_by nodes, so no actual reordering happens.

src/sql/mir.rs
larat7: In `reorder_predicates`, I check `moved_predicates` to avoid adding duplicate predicates. I also use that to avoid adding duplicated filters later on, after inserting group_by nodes.

src/sql/reuse.rs
larat7: Maybe, I'm not seeing something, but these are `or` predicates.  The idea of this test was capturing the `or` example on the appendix of the Finkelstein paper (which, like we discussed, the paper gets it wrong).

src/sql/reuse.rs
larat7: Again, maybe I'm not seeing something, but `a=40` should return `true`.

src/sql/reuse.rs
larat7: Same as above -- `a=40` should return `true`.

src/sql/reuse.rs
ms705: Yeah, I realized that was the goal! The predicate is still always true though, since [-infinity, 80] and [20, infinity] overlap ;-) The `cp1` and `cp2` predicates are the ones that appear in the Finkelstein paper, if I read it correctly -- `cp3` doesn't. But I think it's fine -- the point here is not whether the predicate is always true, but whether the implications are correct (which they are).

src/sql/reuse.rs
ms705: Sorry, my bad, too many predicates :-) `cp1` is just the flipped version of `cp3`; I misread it as being the opposite equality direction (as in `cp3` above)

src/sql/reuse.rs
ms705: Agree, my bad.

src/sql/reuse.rs
larat7: Ah, I see what you mean and now I realize the comments are wrong. They should've been `cp1: a < 20 or a > 60`, `cp2: a < 10 or a > 80` and `cp3: a > 60 or a < 20` (which are the predicates in the paper). Argh, too many predicates.

src/sql/mir.rs
```diff
@@ -785,6 +808,142 @@ impl SqlToMirConverter {
...
+    /// Returns all collumns used in a set of predicates
+    fn cols_in_predicates(
+        &self,
+        ces: Vec<ConditionExpression>
+    ) -> HashSet<Column> {
+        use nom_sql::ConditionExpression::*;
+
+        let mut cols = HashSet::new();
+        for ce in ces {
+            match ce {
```
larat7: Yes. I changed `qgn.predicates` to be a `Vec<ConditionExpression>` rather than a `Vec<ConditionTree>` because it makes it easier to recursively create nodes in `make_predicate_nodes` (there's less `match` statements and so on). But I guess it has the downside of having to repeat some code to get the columns in each predicate.

PR #20, opened 2017-08-04
adds support for nested select statements

larat7:


src/sql/passes/subqueries.rs
```diff
@@ -0,0 +1,223 @@
...
+    q.hash(&mut h);
+    h.finish()
+}
+
+impl SubQueries for SqlQuery {
+    fn extract_and_replace_subqueries(self) -> (SqlQuery, Vec<(String, SqlQuery)>) {
+        match self {
+            SqlQuery::Select(ref st) => {
+                if st.where_clause.is_some() {
+                    let ce = st.clone().where_clause.unwrap();
```
ms705: Is this `clone` required? `st.where_clause.as_ref().unwrap()` might work too, assuming `extract_subqueries_from_condition` can take a `&mut ConditionExpression` rather than an owned `ConditionExpression`

ms705: Alternatively, `st.where_clause.take()` allows extracting from the `where_clause` `Option` and sets it to `None`.

src/sql/passes/subqueries.rs
```diff
@@ -0,0 +1,223 @@
...
+                    let (new_ce, queries) = extract_subqueries_from_condition(ce);
+                    let mut new_st = st.clone();
+                    new_st.where_clause = Some(new_ce);
```
ms705: I think this can be done without cloning the statement: either `take` the where clause earlier and put it back in using `st.where_clause = Some(new_ce)` here, or use `std::mem::replace`.

PR #21, opened 2017-08-28
Add other reuse algorithm options

larat7:
Adds two reuse options other than the `Finkelstein` algorithm: `Relaxed` and `Full`. Also considers multiple reuse candidates, instead of using a scoring function to pick a single one.

- `Relaxed`: This algorithm relaxes some of the constraints in the `Finkelstein` algorithm. More specifically, it doesn't give up if a query graph is not a direct extension of an existing query graph. Instead, it tries to reuse prefixes of each query. 

- `Full`: This algorithm simply considers all query graphs as possible reuse candidates and tries to individually merge each of them into the new query being added. This should yield maximum possible reuse.

- Reuse numbers after migrating `hotsoup` until schema 66. This doesn't count reuse of base tables, queries that are an exact match, or queries that simply added a new reader node.

Reuse Alg  | # Nodes Reused | 
--- | --- |
`Finkelstein` | 34 |
`Relaxed` | 38 |
`Full` | 41 |

### DIFF
--- a/src/sql/reuse/mod.rs
+++ b/src/sql/reuse/mod.rs
@@ -7,6 +7,7 @@ use std::collections::HashMap;
 
 mod finkelstein;
 mod weak;
+mod all;
 mod helpers;
 
 #[derive(Clone, Debug)]
@@ -20,6 +21,7 @@ pub enum ReuseType {
 enum ReuseConfigType {
     Finkelstein,
     Weak,
+    All,
 }
 
 const REUSE_CONFIG: ReuseConfigType = ReuseConfigType::Finkelstein;
@@ -33,13 +35,7 @@ impl ReuseConfig {
         match self.config {
             ReuseConfigType::Finkelstein => finkelstein::Finkelstein::reuse_candidates(qg, query_graphs),
             ReuseConfigType::Weak => weak::Weak::reuse_candidates(qg, query_graphs),
-        }
-    }
-
-    pub fn choose_best_option<'a>(&self, options: Vec<(ReuseType, &'a QueryGraph)>) -> (ReuseType, &'a QueryGraph) {
-         match self.config {
-            ReuseConfigType::Finkelstein => finkelstein::Finkelstein::choose_best_option(options),
-            ReuseConfigType::Weak => weak::Weak::choose_best_option(options)
+            ReuseConfigType::All => all::All::reuse_candidates(qg, query_graphs),
         }
     }
 
@@ -47,6 +43,13 @@ impl ReuseConfig {
         match REUSE_CONFIG {
             ReuseConfigType::Finkelstein => ReuseConfig::finkelstein(),
             ReuseConfigType::Weak => ReuseConfig::weak(),
+            ReuseConfigType::All => ReuseConfig::all(),
+        }
+    }
+
+    pub fn all() -> ReuseConfig {
+        ReuseConfig {
+            config: ReuseConfigType::All,
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/sql/reuse/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    BackjoinRequired(Vec<Table>),
+}
+
+enum ReuseConfigType {
+    Finkelstein,
+    Relaxed,
+    Full,
+}
+
+const REUSE_CONFIG: ReuseConfigType = ReuseConfigType::Finkelstein;
```
ms705: I guess we'll eventually want to turn this into a dynamic option passed when creating a `SqlIncorporator`, rather than a compile-time constant, but this totally fine for now.

src/sql/reuse/relaxed.rs
```diff
@@ -0,0 +1,147 @@
...
+                        _ => return None,
+                    }
+                }
+                QueryGraphEdge::LeftJoin(_) => {
+                    if !new_qg.edges.contains_key(srcdst) { return None; }
+                    let new_qge = &new_qg.edges[srcdst];
+                    match *new_qge {
+                        QueryGraphEdge::LeftJoin(_) => {}
+                        // If there is no matching LeftJoin edge, we cannot reuse
+                        _ => return None,
```
ms705: On that thought (though not relevant as a direct change here, and certainly applicable to all algorithms): we could totally reuse a `LeftJoin` as a plain `Join` by simply adding a filter that discards rows with NULLs in the right side columns. This is neat because filters add no materializations and can be queried through.

src/sql/reuse/relaxed.rs
```diff
@@ -0,0 +1,147 @@
...
+        // 1. NQG's nodes is subset of EQG's nodes
+        // -- already established via signature check
+        assert!(
+            existing_qg
+                .signature()
+                .is_weak_generalization_of(&new_qg.signature())
+        );
+
+        // Check if the queries are join compatible -- if the new query
+        // performs a superset of the joins in the existing query.
```
ms705: Hmm, here's something we might need to be aware of: this code no longer checks for compatibility of join predicates, only that a join between the same two nodes exists. This is correct as long as we only allow equi-joins and as long as join predicates are simple (i.e., no additional ANDs and ORs within them), which I believe is currently always the case.

The old true-to-Finkelstein implementation guaranteed this because it checked all the predicates _first_ before looking at the join edges; here, we rely on the MIR-level reuse algorithm to check correctness of the join columns (which it indeed does).

In other words, no problem as such, but we need to be careful if we allow more powerful joins later (e.g., HotCRP-style `a JOIN b ON (a.c = b.d AND a.c != <something>)`).

src/sql/reuse/relaxed.rs
```diff
@@ -0,0 +1,147 @@
...
+                }
+                if !matched {
+                    // We found no matching predicate for np, so we give up now.
+                    // However, we can still reuse the join nodes from the existing query.
+                    return Some(ReuseType::PrefixReuse);
+                }
+            }
+        }
+
+        // we don't need to check projected columns to reuse a prefix of the query
```
ms705: This comment might need updating in light of the fact that we use "prefix" differently now ;)

src/sql/reuse/relaxed.rs
larat7: Ah, yeah. Ideally, we would also check predicate implication in the the join predicate. Though, I don't think the Finkelstein algorithm actually handles this. It only checks predicate implication in non-edge predicates.

PR #22, opened 2017-09-05
Creates one join node per join predicate

larat7:
This fixes how we handle multiple join predicates for the same pair of `(src, dst)`.

Currently, we don't support joins on multiple columns, so if we have multiple predicates, we need to create one join node per predicate.

This also fixes the issue where this [assertion](https://github.com/mit-pdos/distributary/blob/3e4c4594b971b1bc86a4bcbf676b8774522d5c6e/src/ops/join.rs#L92) was triggered when joining against the same table (by using aliases). So now we can run the following query schema:
```
CREATE TABLE r (a int, b int);
SELECT r1.a as a1, r2.a as a2 from r as r1, r as r2 where r1.a = r2.a;
SELECT r1.a as a1, r2.a as a2 from r as r1, r as r2 where r1.a = r2.b and r2.a = r1.b;
```
However, because we get rid of table aliases before node construction, we might still pick the wrong join columns in the last query, since we can't distinguish between `r1.a` and `r2.a`.

### DIFF
--- a/src/mir/mod.rs
+++ b/src/mir/mod.rs
@@ -1479,31 +1479,44 @@ fn make_join_node(
         n.borrow().column_id_for_column(col)
     };
 
-    let join_config = proj_cols
+    assert_eq!(on_left.len(), 1, "no support for multiple column joins");
+    assert_eq!(on_right.len(), 1, "no support for multiple column joins");
+
+    // this assumes the columns we want to join on appear first in the list
+    // of projected columns. this is fine for joins against different tables
+    // since we assume unique column names in each table. however, this is
+    // not correct for joins against the same table, for example:
+    // SELECT r1.a as a1, r2.a as a2 from r as r1, r as r2 where r1.a = r2.b and r2.a = r1.b;
+    //
+    // the `r1.a = r2.b` join predicate will create a join node with columns: r1.a, r1.b, r2.a, r2,b
+    // however, because the way we deal with alias, we can't distinguish between `r1.a` and `r2.a`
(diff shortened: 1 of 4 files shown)


src/sql/mir.rs
```diff
@@ -1041,39 +1041,55 @@ impl SqlToMirConverter {
...
+                                    jp,
+                                    left_node.clone(),
+                                    prev_join.clone(),
+                                    JoinType::Inner,
+                                );
+
+                                prev_join = cur_join.clone();
+                                new_node_count+=1;
+                                jns.push(cur_join);
+                            }
```
ms705: This code is identical to the one in the `LeftJoin` case, I think? It would be nice not to duplicate it, if it can easily be externalized into a closure. Makes it easier to avoid bugs that would occur if we change one of them but not the other.

src/mir/mod.rs
```diff
@@ -1479,31 +1482,44 @@ fn make_join_node(
...
+
+    // this assumes the columns we want to join on appear first in the list
+    // of projected columns. this is fine for joins against different tables
+    // since we assume unique column names in each table. however, this is
+    // not correct for joins against the same table, for example:
+    // SELECT r1.a as a1, r2.a as a2 from r as r1, r as r2 where r1.a = r2.b and r2.a = r1.b;
+    //
+    // the `r1.a = r2.b` join predicate will create a join node with columns: r1.a, r1.b, r2.a, r2,b
+    // however, because the way we deal with alias, we can't distinguish between `r1.a` and `r2.a`
+    // at this point in the codebase, and because so the `r2.a = r1.b` will join on the wrong `a`
```
ms705: Nit: s/so//

PR #23, opened 2017-09-21
tpc-w benchmark

larat7:


src/flow/migrate/sharding.rs
```diff
@@ -547,6 +547,14 @@ fn reshard(
...
     assert!(!graph[src].is_source());
 
+    if graph[src].sharded_by() == Sharding::None && to == Sharding::None {
+        trace!(log, "no need to shuffle";
+               "src" => ?src,
+               "dst" => ?dst,
+               "sharding" => ?to);
+        return;
+    }
+
```
larat7: @jonhoo is this change supposed to be here?

PR #24, opened 2017-09-26
Use distributary::{Getter, Datas} in basic example

ekmartin:
Couldn't get the example to compile otherwise. Also removed the `basic.srp` schema, since it seemed like the `sql` string in `basic-recipe.rs` replaced it.


jonhoo:
Yeah, we haven't looked at the example code in a while. Your changes look right, though I'd like to get @ms705's opinion on removing `basic.srp` before accepting.

ms705:
Thanks for this! The example was indeed broken, and nobody noticed because it's not part of our normal test suite (which it ought to be -- currently, we have no end-to-end test or benchmark for SQL recipes, actually).

Removing `basic.srp` is fine; the example doesn't use it and the contents are outdated (e.g., we no longer ignore types!).

PR #26, opened 2017-09-26
Web fixes

ekmartin:
A few minor changes related to the web targets:
* Adds `--features=binaries` to the cargo commands in `exercise-web.sh` so that it successfully compiles
* Uses `Recipe::from_str` for the queries in `sql_main.rs`. Previously panicked with a `entered unreachable code: asked for unset index for internal B node', src/flow/node/mod.rs:272:16` which I don't really know what means, but reading in the queries like this seems to have fixed it ¯\\\_(ツ)_/¯
* Prints out the running port number and serves `graph.html` at `/` as well (mostly since I couldn't figure out how it worked initially, and I think this would've made it easier to grok)

Neat graph!

### DIFF
--- a/exercise-web.sh
+++ b/exercise-web.sh
@@ -1,13 +1,13 @@
 #!/bin/sh
 echo "Build"
-cargo build --bin web --features=web
+cargo build --bin web --features=binaries,web
 res=$?
 if [ "$res" -ne 0 ]; then
 	exit "$res"
 fi
 
 echo "Start"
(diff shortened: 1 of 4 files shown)


jonhoo:
LGTM

ms705:
We should chase up the error observed when directly using the `SqlIncorporator` (rather than via `Recipe`), though. I suspect that either we end up with a base node without an index, or with one with an index in the wrong place.

PR #27, opened 2017-09-27
Bump the settle time to get tests to pass on Travis

ekmartin:
Looks like Travis is too slow for the tests in `lib/tests.rs`. Bumping the settle time to 500 ms seems to work (re-ran the tests a few times).


ms705:
Thanks -- we've bumped the time up a few times already, but evidently not enough.

PR #28, opened 2017-09-27
Add basic SQL end to end test

ekmartin:
This adds an end to end test that reads in a table and a query, runs a migration, and finally performs a few queries - all in all pretty similar to `examples/basic-recipe`. 

Context: https://github.com/mit-pdos/distributary/pull/24#issuecomment-332264030

I wanted to add a small test to the actual example as well, but couldn't find a reasonable way to do so - see https://github.com/rust-lang/cargo/issues/3559

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -285,6 +285,41 @@ fn it_works_deletion() {
 }
 
 #[test]
+fn it_works_with_sql_recipe() {
+    let mut g = distributary::Blender::new();
+    let sql = "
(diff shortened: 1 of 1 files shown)


ekmartin:
Seems like `cargo test` is normally supposed to build the examples, but apparently not if the example includes a binary. If we removed `required-features = ["binaries"]` from the example target and instead used `cargo run --example basic-recipe` it would at least get compiled as part of the regular tests. What do you think?

ms705:
Seems like a fine plan to me -- unless @jonhoo has any objections?

The test looks good, and is a useful one to have.

PR #29, opened 2017-09-28
Make migrations happen in a closure

jonhoo:
I've been wanting to make this change for a while now.

Since a migration mutably borrows the `Blender`, the migration usually has to happen in its own (useless) scope, at least without NLL. This is annoying. Furthermore, the requirement to call `mig.commit()` is silly since we (at least atm) cannot abort a migration.

This patch changes migrations to instead happen in a closure, which essentially wraps the same things every caller had to do previously.

### DIFF
--- a/benchmarks/hotsoup/hotsoup.rs
+++ b/benchmarks/hotsoup/hotsoup.rs
@@ -63,66 +63,64 @@ impl Backend {
         use std::fs::File;
 
         let ref blacklist = self.blacklist;
-        // migrate
-        let mut mig = self.g.start_migration();
-
-        let mut sf = File::open(schema_file).unwrap();
-        let mut s = String::new();
-
-        let mut blacklisted = 0;
-
-        // load schema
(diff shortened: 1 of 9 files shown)


ms705:
LGTM :+1: I think it's good that this internally maintains the old way of doing things (i.e., `migrate` still calls `commit`), since that may allow us to bring back the ability to have abortable migrations later.

The only possible complication I can think of is that migrations that now happen inside a closure in a loop may need to capture variables from outside the closure (which can have tricky semantics). However, since that doesn't seem to be the case in any of our examples, we're good, I think.

fintelia:
LGTM, though the NNL pain looks like it will be going away sooner than expected

jonhoo:
@ms705 `migrate` taking an `FnOnce` should make that pretty easy to deal with (since you'll be allowed to mutably borrow, or even move, into the closure).

@fintelia true, though I still think this is a worthwhile change. also, it'll take a while for full NLL to land without the inevitable bugs.

PR #30, opened 2017-10-01
Arithmetic Expressions in Projections

ekmartin:
This adds support for e.g. `SELECT 2 * price FROM ...` and `SELECT price * MAX(column) FROM ...` by handling nom-sql's `ArithmeticExpression`.

Arithmetic between different numeric data types is converted implicitly:
* `BigInt` + `Int` = `BigInt`
* `Real` + `Int` = `Real`
* Not supported: `BigInt` + `Real` (since `Real`'s integer part is represented using an `i32`)

Similar to how nom-sql handles arithmetic this currently only deals with simple, non-nested, expressions, but it should definitely be possible to add support for nested expressions in the future.

I'm not completely sure if the way I've handled `ArithmeticColumns` in the query graph is correct (especially when it comes to re-use), but it seems to work for the tests I've written so far. 
The output formatting of arithmetic nodes could probably also need some tuning.


ms705:
Very nice! I'll have a look over this.

tests/lib.rs
```diff
@@ -320,6 +320,119 @@ fn it_works_with_sql_recipe() {
...
+    assert_eq!(result.len(), 1);
+    assert_eq!(result[0][1], (price as f64 * fraction).into());
+}
+
+#[test]
+fn it_works_with_function_arithmetic() {
+    let mut g = distributary::Blender::new();
+    let sql = "
+        CREATE TABLE Bread (id int, price int, PRIMARY KEY(id));
+        Price: SELECT 2 * COUNT(price) FROM Bread WHERE id = ?;
```
ekmartin: I wanted this query to be `SELECT 2 * MAX(price) FROM Bread` but I couldn't figure out how to retrieve the result without having a key connected to the query. I saw some of the other tests/benchmarks used a bogus key of 0, but `getter.lookup(&0.into(), true)` didn't seem to do the trick either. What am I missing?

ekmartin: I got it working by using the result from `2 * MAX(price)` as the key for now, but would definitely prefer a cleaner way of doing it.

Commit: https://github.com/mit-pdos/distributary/pull/30/commits/688e85fc033e9217369405208714f01c14d7dd71

src/ops/project.rs
```diff
@@ -34,6 +93,25 @@ impl Project {
             self.emit.as_ref().map_or(col, |emit| emit[col])
         }
     }
+
+    fn eval_expression(&self, expression: &ProjectExpression, record: &Record) -> DataType {
+        let left = match expression.left {
+            ProjectExpressionBase::Column(i) => &record[i],
+            ProjectExpressionBase::Literal(ref data) => data,
+        }.clone();
```
jonhoo: I don't think the `.clone()` here (or below) should be necessary?

src/ops/project.rs
```diff
@@ -89,10 +167,20 @@ impl Ingredient for Project {
...
+                }
                 match self.additional {
-                    Some(ref a) => for i in a {
-                        new_r.push(i.clone());
-                    },
+                    Some(ref a) => {
+                        for i in a {
+                            new_r.push(i.clone());
+                        }
+                    }
```
jonhoo: @malte this is a similar case to aggregations where columns appear in a pre-defined location (i.e., first all projected columns, then all expressions, then all literals). I know that affected MIR for aggregations during the deadline, so maybe be on the lookout for that here too?

src/ops/project.rs
ekmartin: Is there any way to do this without overloading the arithmetic operators on `&DataType` instead of `DataType`? Pushed a commit [here](https://github.com/ekmartin/distributary/commit/d1c65298f72a04a9e86ee6966a66f70d3f7f334d0) for how that would look.

Tried playing around with ownership stuff, but couldn't figure out how to do it without having to `.clone()` the `ProjectExpression` passed into `eval_expression` and the `DataType` at `record[i]`.

src/sql/mod.rs
```diff
@@ -720,7 +720,7 @@ mod tests {
             // Should have two nodes: source and "users" base table
             let ncount = mig.graph().node_count();
             assert_eq!(ncount, 2);
-            assert_eq!(get_node(&inc, &*mig, "users").name(), "users");
```
ekmartin: Does the `&*` make a difference in these tests? They seem to be passing fine without it as well, so I removed it - let me know if it should be there though and I'll revert the commit.

ekmartin:
@ms705: I split out the changes from `mir/mod.rs` into the different files you created in https://github.com/mit-pdos/distributary/commit/8590bff9317d46a45d7acf70cec1e5f517c13a51.

ms705:
Thanks, I'll review this tonight or tomorrow at the latest!

src/mir/node.rs
```diff
@@ -772,11 +775,13 @@ impl MirNodeType {
...
                 emit: ref our_emit,
                 literals: ref our_literals,
+                arithmetic: ref our_arithmetic,
             } => match *other {
                 MirNodeType::Project {
                     ref emit,
                     ref literals,
-                } => our_emit == emit && our_literals == literals,
+                    ref arithmetic,
+                } => our_emit == emit && our_literals == literals && our_arithmetic == arithmetic,
```
ms705: This is quite conservative (though probably fine for v0 of arithmetic expression support): it will only allow reuse of projections that have *exactly identical* arithmetic expressions. For example, with multiple elements in the `arithmetic` vector, it would only reuse the projection if the vectors have the same elements and they're all identical. However, it is also acceptable to reuse the projection if the new `our_arithmetic` is a strict subset of `arithmetic`.

Now, you might observe that we are similarly unnecessarily strict about `emit`, and you'd be right ;-)

src/mir/node.rs
```diff
@@ -994,13 +999,26 @@ impl Debug for MirNodeType {
...
+                ref arithmetic,
             } => write!(
                 f,
-                "π [{}{}]",
+                "π [{}{}{}]",
                 emit.iter()
                     .map(|c| c.name.as_str())
                     .collect::<Vec<_>>()
                     .join(", "),
+                if arithmetic.len() > 0 {
```
ms705: nit: `is_empty()`

src/ops/project.rs
```diff
@@ -1,23 +1,82 @@
...
+        left: ProjectExpressionBase,
+        right: ProjectExpressionBase,
+    ) -> ProjectExpression {
+        ProjectExpression {
+            op: op,
+            left: left,
+            right: right,
+        }
+    }
+}
```
ms705: Hmm, I suspect we only need these (as opposed to using `nom-sql`'s almost-identical `ArithmeticExpression`) because of the `Literal(DataType)` in the base case, which in `nom-sql` is `Scalar(Literal)`. That's a bit suboptimal -- could we use the `nom-sql` structure here since we implement `Into<DataType>` for  `nom_sql::Literal`? (I guess we'd potentially be doing the conversion every time we access the literal if we're not careful, but maybe that can be solved without duplicating the data structure.)

src/ops/project.rs
```diff
@@ -1,23 +1,82 @@
...
+
+impl fmt::Display for ProjectExpressionBase {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ProjectExpressionBase::Column(u) => write!(f, "{}", u),
+            ProjectExpressionBase::Literal(ref l) => write!(f, "(lit: {})", l),
```
ms705: Should this just be the value? If I parse this right, we would currently end up with funny-looking output like `salary * (lit: 100)` rather than `salary * 100`.

src/ops/project.rs
ms705: I doing arithmetic on `&DataType` is fine. That said, the problem here is less about cloning the record but about the fact that the whole `ProjectExpressions` is being cloned (which is larger and more complex). Strictly speaking, _only_ `record[i]` needs to be cloned, correct? Then you could use the arithmetic operators directly on `DataType` and `Literal`, without having to overload them for `ProjectExpression`.

src/ops/project.rs
ms705: On further thought, I guess another difference is that the `Column` variant of the `ProjectExpressionBase` enum holds a numeric column index, not a `Column` object. Hrm, maybe we need the duplicate structures after all.

src/sql/mod.rs
ms705: I think this is probably a leftover from when migrations worked somewhat differently, and can indeed be removed.

src/sql/query_graph.rs
```diff
@@ -653,20 +700,23 @@ pub fn to_query_graph(st: &SelectStatement) -> Result<QueryGraph, String> {
...
-                            .entry(String::from("computed_columns"))
-                            .or_insert_with(
-                                || new_node(String::from("computed_columns"), vec![], st),
-                            );
-                        n.columns.push(c.clone());
-                    }
+            FieldExpression::Arithmetic(ref a) => {
+                if let ArithmeticBase::Column(ref c) = a.left {
+                    add_computed_column(&mut qg, c);
+                }
```
ms705: The intention (and if I read this right, the implementation) here is that computed columns (e.g., `SUM(col)`) can occur *within* arithmetic expressions and that we still need to generate the right query graphs for these, correct?

(As opposed to each column defined via an arithmetic expression being considered "computed", which it wouldn't really be from the query graph perspective.)

src/sql/query_graph.rs
```diff
@@ -653,20 +700,23 @@ pub fn to_query_graph(st: &SelectStatement) -> Result<QueryGraph, String> {
...
+
+                if let ArithmeticBase::Column(ref c) = a.right {
+                    add_computed_column(&mut qg, c);
                 }
+
+                qg.columns.push(OutputColumn::Arithmetic(ArithmeticColumn {
+                    name: String::from("arithmetic"),
```
ms705: How does this deal with aliases for arithmetic columns? Seems like it might hardcode the column name always, which would be problematic if we had more than one arithmetic column.

ms705:
A more general thought on this (certainly outside the scope of this PR): I wonder if we need to move to a model where the `Column` object carries information about e.g., arithmetic on projections, or indeed about the column being a literal value. There are two reasons for this:
 1. We already associate information about functions (e.g., `MAX`, `SUM` etc.) with the `Column` object, and it seems inconsistent to treat other similar notions differently.
 2. By carrying information about arithmetic expressions and literals outside the `Column` object, our interfaces keep gaining additional vector of "special" values (most notably the MIR `Project` node and the `Project` operator), and we are forced into a static column emit order -- namely ordinary columns, literal columns, and arithmetic columns in separate groups. This is a little unnecessary, and requires extra projections to permute the columns back into the order the query expects them in; carrying information about arithmetic definitions and literals inside `Column` would permit us to order them any way we like.

The downsides of this plan are the increase in complexity of code handling `Column` objects (and we have plenty of that!) and having to be careful to rewrite columns with arithmetic or literal definitions to "ordinary" columns after their first occurrence (which we're not even doing consistently for function columns currently, but should).

src/ops/project.rs
ekmartin: I added the `lit: ` since it outputs the emit index, and not the actual column name. I thought `0 * 100` might look slightly confusing.

https://github.com/ekmartin/distributary/blob/436333cce567b3c50e253fa44ba9f621c23a576a/src/ops/project.rs#L300

ms705: Ah, fair point. I keep forgetting that the low-level operators only know about column indices...

src/sql/query_graph.rs
ekmartin: Exactly - before I added this it wasn't possible to refer to computed columns inside arithmetic expressions.

src/ops/project.rs
ekmartin: > Strictly speaking, only `record[i]` needs to be cloned, correct? Then you could use the arithmetic operators directly on DataType and Literal, without having to overload them for ProjectExpression.

I might be misunderstanding, but I think that's what I'm doing? `left` and `right` are `DataType` objects here, and the operators are overloaded [directly](https://github.com/mit-pdos/distributary/pull/30/files#diff-712c61e60613c53cdcc824c13afd1800R269) on `DataType`.

src/sql/query_graph.rs
ekmartin: Hm, that's a good point. I implemented this by looking at how it was being done for literals, which also seems to use the same name for the computed columns:
https://github.com/ekmartin/distributary/blob/85044bf28c9bb83bbe2b907bfdfaf7cf0b6d0c84/src/sql/query_graph.rs#L696-L702


Both literals and arithmetic expression names are handled here though:
https://github.com/ekmartin/distributary/blob/85044bf28c9bb83bbe2b907bfdfaf7cf0b6d0c84/src/sql/mir.rs#L737-L763

Where column objects are created with only the name and the table. It doesn't seem to look at aliases at all for literals though, which I guess might mean that aliases doesn't work for literals either?

I added a test for multiple arithmetic expressions in one query [here](https://github.com/mit-pdos/distributary/pull/30/commits/28a9dadb8743cc9d18e19b7140ac83a8fdee5b39) - I'll see how it behaves with aliases as well.

tests/lib.rs
```diff
@@ -323,6 +323,149 @@ fn it_works_with_sql_recipe() {
...
+    let price: distributary::DataType = 123.into();
+    mutator.put(vec![id.clone(), price]).unwrap();
+
+    // Let writes propagate:
+    sleep();
+
+    // Retrieve the result of the count query:
+    let result = getter.lookup(&id, true).unwrap();
+    assert_eq!(result.len(), 1);
+    assert_eq!(result[0][1], 246.into());
```
ekmartin: I think I might have gotten something wrong with the output order here. For other queries it looks like the query key is outputted as the [last element in the result](https://github.com/ekmartin/distributary/blob/28a9dadb8743cc9d18e19b7140ac83a8fdee5b39/tests/lib.rs#L320-L322) - whereas with arithmetic expressions it seems to be placed first, which I think might be wrong?

ekmartin:
@ms705: Sorry about the late answer. Storing the information on `Column` seems more reasonable indeed. That would also make it easier to add support for aliases in arithmetic columns in nom-sql, as it wouldn't need to be tacked onto e.g. both `Literal` and `ArithmeticExpression`.

Maybe it would be possible to move the translation from columns to column IDs from [to_flow#make_project_node](https://github.com/mit-pdos/distributary/blob/86c0461cf2ce16d6f9e73816ea762ca0e4257b15/src/mir/to_flow.rs#L374-L376) to the `Project` operator itself? That way the full list of columns in the right order could be attached to the operator, and something like `parent.column_id_for_column()` could be done from there.

tests/lib.rs
ekmartin: Never mind - this seems to match the output order for literals (and I think it makes sense since the columns are just outputted in groups like mentioned in https://github.com/mit-pdos/distributary/pull/30#issuecomment-334186613)

src/sql/query_graph.rs
ms705: :+1:

src/sql/query_graph.rs
ms705: Yeah, you're right that aliases for literals probably don't work.

Maybe it's best to add alias support to both `ArithmeticColumn` and `LiteralColumn` for now (either within this PR or separately, I don't mind).

src/ops/project.rs
ms705: Oh, yes, you're right, I misread.

ms705:
Are we good to merge this? @jonhoo @ekmartin

jonhoo:
I still think it's sad that we `clone` the `DataType`, but I guess it's fine since it'll always be numbers. So yeah, I'm fine with merging.

ms705:
@jonhoo Agreed; do you see a way to do this without a clone? As far as I can see, arithmetic on `&DataType` is the only option, right?

ms705:
@ekmartin We had an offline discussion here and concluded that arithmetic on `&DataType` is the best solution. Can you push that commit into this PR?

ekmartin:
@ms705: Sounds good. Had to make a [new PR](https://github.com/mit-pdos/distributary/pull/35), since the last commit wouldn't show up in this one for some reason (still doesn't).

Might be a good idea to push the branch to `mit-pdos/distributary` as well before merging, so that taster runs the benchmarks?

PR #31, opened 2017-10-02
Run rustfmt on Travis

ekmartin:
This runs `rustfmt` with `--write-mode=diff` on Travis, so that the build [fails](https://travis-ci.org/ekmartin/distributary/builds/282444725) if there are formatting errors. It installs `rustfmt` from git, with the SHA `6e41100725267974fa6dcc61134d4377b676ad01`.

Also shortened a few lines that were above 100 characters, to make the build pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,9 @@ rust:
 cache: cargo
 env:
   - SETTLE_TIME=2000
+before_script:
+  - export PATH="$PATH:$HOME/.cargo/bin"
+  - which rustfmt || cargo install --git https://github.com/rust-lang-nursery/rustfmt.git --rev 6e41100725267974fa6dcc61134d4377b676ad01
+script:
+  - cargo fmt -- --write-mode=diff
+  - cargo test --verbose
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
   - SETTLE_TIME=2000
 before_script:
   - export PATH="$PATH:$HOME/.cargo/bin"
-  - which rustfmt || cargo install --git https://github.com/rust-lang-nursery/rustfmt.git --rev 6e41100725267974fa6dcc61134d4377b676ad01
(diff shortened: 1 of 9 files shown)


jonhoo:
Good idea. I'm guessing we should rebase this on top of #32 to see that it actually works.

ekmartin:
Yep! Although it did pass here earlier: https://travis-ci.org/ekmartin/distributary/builds/282452240

jonhoo:
@ekmartin merged #32 -- rebase?

ekmartin:
Rebased :)

ekmartin:
Should be good to go now @jonhoo (added a commit for two more long lines: https://github.com/mit-pdos/distributary/pull/31/commits/b7d7fa9444afe63ff6c19510a986df46fd385a4e)

jonhoo:
@ekmartin why did you close this?

ekmartin:
Woops, sorry. Deleted a bunch of branches on my fork while testing stuff with taster.

ekmartin:
`cargo fmt` [failed with a weird error](https://travis-ci.org/ekmartin/distributary/builds/282528253#L465) when using the cached `rustfmt` binary so removed the `which rustfmt ||` clause in https://github.com/mit-pdos/distributary/pull/31/commits/e2b5649f500c0d7e49538cde626da5bb7fef22d5 so it reinstalls each time instead.

`rustfmt: error while loading shared libraries: libsyntax-d36743e8d4fe92fe.so: cannot open shared object file: No such file or directory`

ms705:
IIRC, this happens if the Rust nightly is updated without also reinstalling `rustfmt`; we've had to locally do `cargo install -f --git https://github.com/rust-lang-nursery/rustfmt --rev <revision>` after each Rust upgrade. The reason is that `rustfmt` depends on a matching `libsyntax` from the Rust compiler.

I suspect something similar happens on Travis (e.g., they cache an old version of `rustfmt`).

jonhoo:
@ms705 is right -- this is caused by a version mismatch between the compiler and rustfmt. I've cleared the travis cache, which tends to help, but it's annoying to deal with nonetheless.

PR #32, opened 2017-10-02
Sleep for longer in tests only on Travis

ekmartin:
Seems like tests like `it_works_basic` still fails on Travis sometimes with 500 ms sleep time. This adds a `SETTLE_TIME` environment variable, so that the sleep time can be increased for Travis without slowing down regular test runs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,5 @@ language: rust
 rust:
   - nightly
 cache: cargo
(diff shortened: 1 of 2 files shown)


jonhoo:
Nice! Though can we then also reduce the `SETTLE_TIME` for non-travis runs?

PR #34, opened 2017-10-04
Fix example error

ekmartin:
This fixes a compile error in the example from the change to `.migrate` closures (`return Err(e)` would be a return for the closure, not the entire `load_recipe` function).

Also removes `required-features = ["binaries"]` from the example to make `cargo test` check it for compile errors, and adds a small note to the README about `cargo run --example basic-recipe`.

### DIFF
--- a/examples/basic-recipe.rs
+++ b/examples/basic-recipe.rs
@@ -6,7 +6,7 @@ use std::{thread, time};
 use distributary::{Blender, Recipe};
 use backend::Backend;
 
-fn load_recipe() -> Result<Backend, String> {
+fn load_recipe() -> Backend {
     // inline recipe definition
     let sql = "# base tables
                CREATE TABLE Article (aid int, title varchar(255), \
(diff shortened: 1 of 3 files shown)


examples/basic-recipe.rs
```diff
@@ -26,21 +26,17 @@ fn load_recipe() -> Result<Backend, String> {
...
         // install recipe
-        match Recipe::from_str(&sql, None) {
-            Ok(mut recipe) => {
-                recipe.activate(mig, false)?;
-                recipe
-            }
-            Err(e) => return Err(e),
-        }
+        let mut recipe = Recipe::from_str(&sql, None).unwrap();
+        recipe.activate(mig, false).unwrap();
```
ekmartin: I changed this to using `.unwrap()` as I think it makes the example more readable. Let me know if you disagree and I'll change it back to returning a `Result`

jonhoo:
LGTM.

PR #36, opened 2017-10-21
Upgrade timekeeper to 0.3.0

ekmartin:
Also removes the macOS comment, since that now works out of the box.


fintelia:
Looks good to me.

PR #37, opened 2017-10-22
Recovery

ekmartin:
This adds a `Blender.recover()` method that goes through and replays the log entries it finds for its base nodes. 

* The log flushes are now separated by a newline, so that each JSON array can be parsed individually.
* `persistence::Parameters` now has an optional `log_prefix` that'll be used to replace `soup` in e.g. `soup-log_0_0-0.json`. Without this parallel tests would sometimes log to the same file - at the same time.
* Packets are re-created from the log entries and passed on to `Domain.handle`. Transactional packets are first passed through `checktable.apply_batch` to retrieve a timestamp. I'm not sure if the latter is completely correct though - cc @fintelia.

I've added a couple of tests, but let me know if there's anything else I should cover.

I also changed the basic example so that it calls `.recover()` and votes with a unique timestamp (so that the count increments each time it's run). Let me know if I should move it to another example or something though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ binaries = ["default"]
 generate_mysql_tests = ["default"]
 
 [dependencies]
-arccstr = "1.0.2"
+arccstr = "1.0.3"
 arrayvec = "0.4.0"
 bincode = "0.9.0"
 buf_redux = "0.6.1"
@@ -78,6 +78,7 @@ default-features = false
 backtrace = { version = "0.3.2", features = ["serialize-serde"] }
 toml = "0.4.1"
 diff = "0.1.10"
+time = "0.1.38"
 
 [profile.release]
 debug=true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,7 @@ backtrace = { version = "0.3.2", features = ["serialize-serde"] }
 toml = "0.4.1"
 diff = "0.1.10"
 time = "0.1.38"
+glob = "0.2.11"
 
 [profile.release]
 debug=true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,6 @@ default-features = false
 backtrace = { version = "0.3.2", features = ["serialize-serde"] }
 toml = "0.4.1"
 diff = "0.1.10"
-time = "0.1.38"
 glob = "0.2.11"
 
 [profile.release]
--- a/dataflow/src/checktable/mod.rs
+++ b/dataflow/src/checktable/mod.rs
@@ -6,7 +6,7 @@
 
 use petgraph::graph::NodeIndex;
 
-use std::collections::HashMap;
+use std::collections::hash_map::{Entry, HashMap};
 use std::fmt;
 use std::fmt::Debug;
 
@@ -283,6 +283,40 @@ impl CheckTable {
                 }
             }
         }
+    }
+
+    pub fn recover(
+        &mut self,
+        base: NodeIndex,
+        columns: usize,
+    ) -> (i64, Option<Box<HashMap<domain::Index, i64>>>) {
+        // Take timestamp
+        let ts = self.next_timestamp;
+        self.next_timestamp += 1;
+
+        // Compute the previous timestamp that each domain will see before getting this one
+        let prev_times = self.compute_previous_timestamps(Some(base));
+
+        // Update checktables
+        self.last_base = Some(base);
+        self.toplevel.insert(base, ts);
+
+        // Set the timestamp for each column:
+        let t = &mut self.granular.entry(base).or_default();
+        for i in 0..columns {
+            match t.entry(i) {
+                Entry::Occupied(mut entry) => {
+                    let g = entry.get_mut();
+                    assert!(g.0.is_empty(), "checktable should be empty before recovery");
+                    g.1 = ts;
+                }
+                Entry::Vacant(entry) => {
+                    entry.insert((HashMap::new(), ts));
+                }
+            }
+        }
+
+        (ts, prev_times)
     }
 
     pub fn apply_unconditional(
--- a/dataflow/src/checktable/mod.rs
+++ b/dataflow/src/checktable/mod.rs
@@ -285,6 +285,8 @@ impl CheckTable {
         }
     }
 
+    // Reserve a timestamp for the given base node, and update each column to said timestamp.
+    // This should be called for each batch of recovery updates.
     pub fn recover(&mut self, base: NodeIndex) -> (i64, Option<Box<HashMap<domain::Index, i64>>>) {
         // Take timestamp
         let ts = self.next_timestamp;
--- a/dataflow/src/checktable/mod.rs
+++ b/dataflow/src/checktable/mod.rs
@@ -6,7 +6,7 @@
 
 use petgraph::graph::NodeIndex;
 
-use std::collections::hash_map::{Entry, HashMap};
+use std::collections::HashMap;
 use std::fmt;
 use std::fmt::Debug;
 
@@ -285,11 +285,7 @@ impl CheckTable {
         }
     }
 
-    pub fn recover(
-        &mut self,
-        base: NodeIndex,
-        columns: usize,
-    ) -> (i64, Option<Box<HashMap<domain::Index, i64>>>) {
+    pub fn recover(&mut self, base: NodeIndex) -> (i64, Option<Box<HashMap<domain::Index, i64>>>) {
         // Take timestamp
         let ts = self.next_timestamp;
         self.next_timestamp += 1;
@@ -301,19 +297,10 @@ impl CheckTable {
         self.last_base = Some(base);
         self.toplevel.insert(base, ts);
 
-        // Set the timestamp for each column:
         let t = &mut self.granular.entry(base).or_default();
-        for i in 0..columns {
-            match t.entry(i) {
-                Entry::Occupied(mut entry) => {
-                    let g = entry.get_mut();
-                    assert!(g.0.is_empty(), "checktable should be empty before recovery");
-                    g.1 = ts;
-                }
-                Entry::Vacant(entry) => {
-                    entry.insert((HashMap::new(), ts));
-                }
-            }
+        for (_column, g) in t.iter_mut() {
+            assert!(g.0.is_empty(), "checktable should be empty before recovery");
+            g.1 = ts;
         }
 
         (ts, prev_times)
--- a/dataflow/src/domain/mod.rs
+++ b/dataflow/src/domain/mod.rs
@@ -5,7 +5,6 @@ use std::thread;
 use std::time;
 use std::io::{BufRead, BufReader};
 use std::fs::File;
-use std::path::PathBuf;
 use std::collections::hash_map::Entry;
 use std::rc::Rc;
 
@@ -1579,14 +1578,8 @@ impl Domain {
             .map(|(index, _node)| index)
             .collect::<Vec<_>>();
 
-        for index in indices {
-            let path = PathBuf::from(&format!(
-                "soup-log-{}_{}-{}.json",
-                self.index.index(),
-                self.shard.unwrap_or(0),
-                index.id(),
-            ));
-
+        for node_index in indices {
+            let path = persistence::log_path(&node_index, self.index, self.shard.unwrap_or(0));
             if !path.exists() {
                 debug!(
                     self.log,
@@ -1609,7 +1602,7 @@ impl Domain {
                 .for_each(|records| {
                     let packet = box Packet::Message {
                         data: records,
-                        link: Link::new(index, index),
+                        link: Link::new(node_index, node_index),
                         tracer: None,
                     };
 
--- a/dataflow/src/domain/mod.rs
+++ b/dataflow/src/domain/mod.rs
@@ -1575,11 +1575,20 @@ impl Domain {
     fn handle_recovery(&mut self) {
         let indices = self.nodes
             .iter()
-            .map(|(index, _node)| index)
+            .map(|(_index, node)| {
+                let node = node.borrow();
+                (node.local_addr().clone(), node.global_addr().clone())
+            })
             .collect::<Vec<_>>();
 
-        for node_index in indices {
-            let path = persistence::log_path(&node_index, self.index, self.shard.unwrap_or(0));
+        for (local_addr, global_addr) in indices {
+            let path = persistence::log_path(
+                &local_addr,
+                self.index,
+                self.shard.unwrap_or(0),
+                self.persistence_parameters.log_prefix.clone(),
+            );
+
             if !path.exists() {
                 debug!(
                     self.log,
@@ -1590,20 +1599,37 @@ impl Domain {
                 continue;
             }
 
+            // TODO(ekmartin): move to GroupCommitQueue
             debug!(self.log, "Recovering from file"; "path" => format!("{:?}", path));
             let file = File::open(path).unwrap();
             BufReader::new(file)
                 .lines()
                 .flat_map(|line| {
                     let line = line.unwrap();
-                    let records: Vec<Records> = serde_json::from_str(&line).unwrap();
+                    let records: Vec<persistence::LogEntry> = serde_json::from_str(&line).unwrap();
                     records
                 })
-                .for_each(|records| {
-                    let packet = box Packet::Message {
-                        data: records,
-                        link: Link::new(node_index, node_index),
-                        tracer: None,
+                .for_each(|entry| {
+                    let packet = match entry.packet_type {
+                        persistence::PacketType::Message => box Packet::Message {
+                            data: entry.records,
+                            link: Link::new(local_addr, local_addr),
+                            tracer: None,
+                        },
+                        persistence::PacketType::Transaction => {
+                            let (ts, prev) = self.transaction_state
+                                .get_checktable()
+                                .lock()
+                                .unwrap()
+                                .apply_unconditional(global_addr, &entry.records);
+
+                            box Packet::Transaction {
+                                data: entry.records,
+                                link: Link::new(local_addr, local_addr),
+                                state: TransactionState::Committed(ts, global_addr, prev),
+                                tracer: None,
+                            }
+                        }
                     };
 
                     self.handle(packet);
--- a/dataflow/src/domain/mod.rs
+++ b/dataflow/src/domain/mod.rs
@@ -1573,6 +1573,13 @@ impl Domain {
     }
 
     fn handle_recovery(&mut self) {
+        let group_commit_queues = persistence::GroupCommitQueueSet::new(
+            self.index,
+            self.shard.unwrap_or(0),
+            &self.persistence_parameters,
+            self.transaction_state.get_checktable().clone(),
+        );
+
         let indices = self.nodes
             .iter()
             .map(|(_index, node)| {
@@ -1582,13 +1589,7 @@ impl Domain {
             .collect::<Vec<_>>();
 
         for (local_addr, global_addr) in indices {
-            let path = persistence::log_path(
-                &local_addr,
-                self.index,
-                self.shard.unwrap_or(0),
-                self.persistence_parameters.log_prefix.clone(),
-            );
-
+            let path = group_commit_queues.log_path(&local_addr);
             if !path.exists() {
                 debug!(
                     self.log,
--- a/dataflow/src/domain/mod.rs
+++ b/dataflow/src/domain/mod.rs
@@ -3,8 +3,6 @@ use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time;
-use std::io::{BufRead, BufReader};
-use std::fs::File;
 use std::collections::hash_map::Entry;
 use std::rc::Rc;
 
@@ -20,7 +18,6 @@ use transactions;
 use persistence;
 use debug;
 use checktable;
-use serde_json;
 use slog::Logger;
 use timekeeper::{RealTime, SimpleTracker, ThreadTime, Timer, TimerSet};
 use tarpc::sync::client::{self, ClientExt};
@@ -1580,62 +1577,10 @@ impl Domain {
             self.transaction_state.get_checktable().clone(),
         );
 
-        let indices = self.nodes
-            .iter()
-            .map(|(_index, node)| {
-                let node = node.borrow();
-                (node.local_addr().clone(), node.global_addr().clone())
-            })
-            .collect::<Vec<_>>();
-
-        for (local_addr, global_addr) in indices {
-            let path = group_commit_queues.log_path(&local_addr);
-            if !path.exists() {
-                debug!(
-                    self.log,
-                    "Could not find file while recovering";
-                    "path" => format!("{:?}", path)
-                );
-
-                continue;
-            }
-
-            // TODO(ekmartin): move to GroupCommitQueue
-            debug!(self.log, "Recovering from file"; "path" => format!("{:?}", path));
-            let file = File::open(path).unwrap();
-            BufReader::new(file)
-                .lines()
-                .flat_map(|line| {
-                    let line = line.unwrap();
-                    let records: Vec<persistence::LogEntry> = serde_json::from_str(&line).unwrap();
-                    records
-                })
-                .for_each(|entry| {
-                    let packet = match entry.packet_type {
-                        persistence::PacketType::Message => box Packet::Message {
-                            data: entry.records,
-                            link: Link::new(local_addr, local_addr),
-                            tracer: None,
-                        },
-                        persistence::PacketType::Transaction => {
-                            let (ts, prev) = self.transaction_state
-                                .get_checktable()
-                                .lock()
-                                .unwrap()
-                                .apply_unconditional(global_addr, &entry.records);
-
-                            box Packet::Transaction {
-                                data: entry.records,
-                                link: Link::new(local_addr, local_addr),
-                                state: TransactionState::Committed(ts, global_addr, prev),
-                                tracer: None,
-                            }
-                        }
-                    };
-
-                    self.handle(packet);
-                });
-        }
+        group_commit_queues
+            .retrieve_recovery_packets(&self.nodes)
+            .into_iter()
+            .for_each(|packet| self.handle(packet));
 
         self.control_reply_tx
             .as_ref()
--- a/dataflow/src/domain/mod.rs
+++ b/dataflow/src/domain/mod.rs
@@ -3,6 +3,9 @@ use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time;
+use std::io::{BufRead, BufReader};
+use std::fs::File;
+use std::path::PathBuf;
 use std::collections::hash_map::Entry;
 use std::rc::Rc;
 
@@ -18,6 +21,7 @@ use transactions;
 use persistence;
 use debug;
 use checktable;
+use serde_json;
 use slog::Logger;
 use timekeeper::{RealTime, SimpleTracker, ThreadTime, Timer, TimerSet};
 use tarpc::sync::client::{self, ClientExt};
@@ -803,6 +807,9 @@ impl Domain {
             Packet::ReplayPiece { .. } => {
                 self.handle_replay(m);
             }
+            Packet::StartRecovery { .. } => {
+                self.handle_recovery();
+            }
             consumed => {
                 match consumed {
                     // workaround #16223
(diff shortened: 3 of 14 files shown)


src/flow/persistence/mod.rs
```diff
@@ -253,6 +286,64 @@ impl GroupCommitQueueSet {
...
+                        },
+                        PacketType::Transaction => {
+                            let id = checktable::TransactionId(i as u64);
+                            let transactions = vec![(id, data.clone(), None)];
+                            let request = checktable::service::TimestampRequest {
+                                transactions,
+                                base: global_addr,
+                            };
+
+                            let reply = self.checktable.apply_batch(request).unwrap().unwrap();
```
ekmartin: Not sure if this should handle `None` as well, like in `merge_transactional_packets`. I thought maybe it didn't have to since all the transactions are already committed, but that might be a false assumption.

src/flow/mod.rs
```diff
@@ -295,6 +296,22 @@ impl Blender {
...
+        info!(self.log, "Recovering from log");
+        let indices = self.inputs()
+            .iter()
+            .map(|&(_input, ref node)| node.domain())
+            .collect::<Vec<_>>();
+
+        for index in indices {
+            let domain = self.domains.get_mut(&index).unwrap();
+            domain.send(box payload::Packet::StartRecovery).unwrap();
+            domain.wait_for_ack().unwrap();
```
fintelia: Is it necessary to have domains recover sequentially? Could they all do so in parallel?

src/flow/persistence/mod.rs
```diff
@@ -191,21 +215,30 @@ impl GroupCommitQueueSet {
...
                 {
-                    let data_to_flush: Vec<_> = self.pending_packets[&node]
+                    let entries_to_flush: Vec<_> = self.pending_packets[&node]
                         .iter()
                         .map(|p| match **p {
-                            Packet::Transaction { ref data, .. } |
-                            Packet::Message { ref data, .. } => data,
+                            Packet::Transaction { ref data, .. } => LogEntry {
+                                records: Cow::Borrowed(data),
+                                packet_type: PacketType::Transaction,
```
fintelia: I don't think it is necessary to store the packet type on a per message basis. All writes to a base node should be of the same type (and you can check which it is by seeing if the base node is transactional)

src/flow/persistence/mod.rs
fintelia: It should not be possible to get back `None` since the packet doesn't have a token (and thus is guaranteed to commit).

However, we may not want to use apply_batch since it is actually a somewhat expensive operation. In this specific case, it should be possible to write something simpler that doesn't painstakingly simulate all the granular checktables since the beginning of time. Plus, then you wouldn't have to deal with the TransactionId stuff.

src/flow/persistence/mod.rs
```diff
@@ -253,6 +286,64 @@ impl GroupCommitQueueSet {
...
+        for (_index, node) in nodes.iter() {
+            let node = node.borrow();
+            let local_addr = node.local_addr();
+            let global_addr = node.global_addr();
+            let path = self.log_path(&local_addr);
+            if !path.exists() {
+                continue;
+            }
+
+            let file = File::open(path).unwrap();
```
fintelia: Instead of checking whether the file exists and then unwrapping, it may be better to just try opening the log and catching any errors that triggers.

src/flow/persistence/mod.rs
ekmartin: Ah, didn't think about that. Good point!

src/flow/mod.rs
ekmartin: Yep!

src/flow/persistence/mod.rs
ekmartin: I used `checktable.apply_unconditional` originally - not sure if that's better?

ekmartin:
Regarding `GroupCommitQueueSet`: I'm not really using any of the merge methods there, I just put the retrieval method on it as they shared a few of the struct's fields - I'll pull it out into a standalone function since they're not really related, as you're saying.

src/flow/persistence/mod.rs
```diff
@@ -66,13 +75,100 @@ impl Parameters {
...
+            Ok(f) => f,
+            Err(ref e) if e.kind() == ErrorKind::NotFound => continue,
+            Err(e) => panic!("Could not open log file {:?}: {}", path, e),
+        };
+
+        BufReader::new(file)
+            .lines()
+            .flat_map(|line| {
+                let line = line.unwrap();
+                let entries: Vec<Records> = serde_json::from_str(&line).unwrap();
```
fintelia: If the system was part way through writing a line when it crashed, then deserialization may fail during recovery. It is worth handling this case.

src/flow/persistence/mod.rs
```diff
@@ -66,13 +75,100 @@ impl Parameters {
...
+                let link = Link::new(*local_addr, *local_addr);
+                if node.is_transactional() {
+                    let id = checktable::TransactionId(i as u64);
+                    let transactions = vec![(id, data.clone(), None)];
+                    let request = checktable::service::TimestampRequest {
+                        transactions,
+                        base: global_addr,
+                    };
+
+                    let reply = checktable.apply_batch(request).unwrap().unwrap();
```
fintelia: It would be better to avoid getting a separate timestamp for every packet. Instead, merge packets from the same base node and assign the combination a single timestamp. Based on other experiments, ~512 packets to a batch will probably provide good efficiency.

src/flow/domain/mod.rs
```diff
@@ -1502,6 +1505,21 @@ impl Domain {
...
+    fn handle_recovery(&mut self) {
+        let packets = persistence::retrieve_recovery_packets(
+            &self.nodes,
+            self.index,
+            self.shard.unwrap_or(0),
+            &self.persistence_parameters,
+            self.transaction_state.get_checktable().clone(),
+        );
+
+        packets.into_iter().for_each(|packet| self.handle(packet));
```
fintelia: It is kind of wasteful to construct a Vec if you are going to immediately call into_iter on it. Instead, you could just return an iterator of all the packets.

src/flow/persistence/mod.rs
ekmartin: Sounds good. I guess it doesn't hurt to merge regular packets as well, and not just transactional ones?

src/flow/domain/mod.rs
ekmartin: Agreed - but I couldn't find a clean way of returning an iterator because of the borrow on `node`. Tried something like this:

```rust
    nodes
        .iter()
        .filter_map(|(_index, node)| {
            let node = node.borrow();
            let local_addr = node.local_addr();
            let global_addr = node.global_addr();
            let path = params.log_path(&local_addr, domain_index, domain_shard);
            let file = match File::open(&path) {
                Ok(f) => f,
                Err(ref e) if e.kind() == ErrorKind::NotFound => return None,
                Err(e) => panic!("Could not open log file {:?}: {}", path, e),
            };

            let iter = BufReader::new(file)
                .lines()
                .filter_map(|line| {
                    let line = line.unwrap();
                    let entries: Result<Vec<Records>, _> = serde_json::from_str(&line);
                    entries.ok()
                })
                .flat_map(|r| r)
                // Merge packets into batches of RECOVERY_BATCH_SIZE:
                .chunks(RECOVERY_BATCH_SIZE)
                .into_iter()
                .map(|chunk| chunk.fold(Records::default(), |mut acc, ref mut data| {
                    acc.append(data);
                    acc
                }))
                // Then create Packet objects from the data:
                .map(|data| {
                    let link = Link::new(*local_addr, *local_addr);
                    if node.is_transactional() {
                        let (ts, prevs) = checktable
                            .recover(global_addr, node.fields().len())
                            .unwrap();

                        box Packet::Transaction {
                            link,
                            data,
                            tracer: None,
                            state: TransactionState::Committed(ts, global_addr, prevs),
                        }
                    } else {
                        box Packet::Message {
                            link,
                            data,
                            tracer: None,
                        }
                    }
                });

            Some(iter)
        })
        .flat_map(|i| i)
```

src/flow/persistence/mod.rs
ekmartin: Done!

src/flow/persistence/mod.rs
ekmartin: Fixed in 57cd580f1b75406e40c837e832058bbb5cf87fed

src/checktable/mod.rs
```diff
@@ -285,6 +285,27 @@ impl CheckTable {
...
+
+        // Compute the previous timestamp that each domain will see before getting this one
+        let prev_times = self.compute_previous_timestamps(Some(base));
+
+        // Update checktables
+        self.last_base = Some(base);
+        self.toplevel.insert(base, ts);
+
+        let t = &mut self.granular.entry(base).or_default();
+        for (_column, g) in t.iter_mut() {
```
ekmartin: @fintelia: I updated this so it only sets timestamps for tracked columns, like we talked about.

src/flow/domain/mod.rs
fintelia: I see the issue. Could something like the following work?

```rust
fn handle_recovery(&mut self) {
    let addresses: Vec<_> = nodes.iter().map(|n| (n.local_addr(), n.global_addr())).collect();
    for (local, global) in addresses {
         for packet in persistence::retrieve_recovery_packets(...) {
              self.handle(packet);
         }
    }
}
```

ekmartin: @fintelia: I ended up moving the `retrieve_recovery_packets` logic up to `handle_recovery` - some of the arguments ended up looking a bit weird, so not sure if it was a very good abstraction anyway. Let me know if you disagree though and I'll take a second look at it.

For reference I also bumped into a separate issue when I tried returning an iterator from `retrieve_recovery_packets`: `.chunks()` returns `IntoChunks` - [not an iterable](https://docs.rs/itertools/0.7.2/itertools/trait.Itertools.html#method.chunks) which meant that the value [here](https://gist.github.com/ekmartin/03569ec5f2d8bde35efcac918d71ef44#file-persistence-rs-L17) didn't last as long as `'a`.

fintelia:
@ekmartin There was just a rather large refactor that landed on master. Could you try to rebase/merge the changes into your branch? A lot of things moved around so let me know if you need a hand finding something. Notable moves:

src/flow/domain/mod.rs -> dataflow/src/domain/mod.rs
src/flow/persistence/mod.rs -> dataflow/src/persistence/mod.rs
src/checktable -> dataflow/src/checktable
src/flow/payload.rs -> dataflow/src/payload.rs
tests/lib.rs -> src/tests.rs

ekmartin:
@fintelia: I rebased and added a `.recover()` RPC command to Blender. Went over the tests quickly and updated them so that they pass again, but I can take a more thorough look at it tomorrow if there's anything else that should change with the refactor.

dataflow/src/domain/mod.rs
```diff
@@ -1507,6 +1516,71 @@ impl Domain {
...
+                .filter_map(|line| {
+                    let line = line.unwrap();
+                    let entries: Result<Vec<Records>, _> = serde_json::from_str(&line);
+                    entries.ok()
+                })
+                .flat_map(|r| r)
+                // Merge packets into batches of RECOVERY_BATCH_SIZE:
+                .chunks(RECOVERY_BATCH_SIZE)
+                .into_iter()
+                .map(|chunk| chunk.fold(Records::default(), |mut acc, ref mut data| {
```
ms705: OOI, why is this step required? AIUI, the original `Vec<Records>` exists because each line can contain several records (one originally-written batch per line?), so the real type is `Vec<Vec<Vec<DataType>>>`. The `flat_map` removes the outermost layer, so we now basically have a single, large `Vec<Vec<DataType>>`, which we then chunk into batches. This turns into a new iterator over these batched chunks (iterator over type `&[Vec<DataType>]`, within which each chunk is concerted into a single `Records` structure (= `Vec<Vec<DataType>>`). Is that correct?

(In other words, the extra `map`/`fold` is required to deal with the slice type and get back to `Records`.)

dataflow/src/domain/mod.rs
```diff
@@ -1507,6 +1516,71 @@ impl Domain {
...
+            let checktable = self.transaction_state.get_checktable();
+            let path = self.persistence_parameters.log_path(
+                &local_addr,
+                self.index,
+                self.shard.unwrap_or(0),
+            );
+
+            let file = match File::open(&path) {
+                Ok(f) => f,
+                Err(ref e) if e.kind() == ErrorKind::NotFound => continue,
```
ms705: If the log file doesn't exist, we skip recovering this node -- is this the right behavior? It seems like you could easily end up with a partially recovered database. Perhaps we should at least print a warning here.

ms705:
Looks good! A few questions to check that I got the right understanding:

 1) This currently assumes that the log files for all nodes in a domain exist locally wherever that domain executes (seems reasonable).

 2) While handling a recovery message, a domain does not process normal writes. However, if writes arrived to the domain *before* the `StartRecovery` control packet arrived, there might be some writes stored already -- for a transactional base, we'd crash (because the check table isn't empty), while for a non-transactional one, everything should work out fine post-recovery since the new writes commute with the recovered ones. 

 Some funky subtleties might exist if the write that raced ahead of the recovery is a deletion for something that exists in the to-be-recovered state. In practice, this can't happen currently, because a restart/recovery invalidates client mutators and getters (thus no deletions can be in flight), and because a post-restart deletion racing ahead of the recovery would fail on an empty post-restart base node.

dataflow/src/domain/mod.rs
ekmartin: After taking a second look at this I'm pretty sure that the original implementation is actually wrong, thanks for asking about it.

Since it only called `flat_map` once, it would essentially chunk based on `Records` - and not `Record`:

```rust
.filter_map(|line| {
    let line = line.unwrap();
    let entries: Result<Vec<Records>, _> = serde_json::from_str(&line);
    entries.ok()
})
// At this point we have an iterator over `Vec<Records>` (lines in the log)
.flat_map(|r| r)
// We've now gone from an iterator over `Vec<Records>` to an iterator over `Records`
.chunks(...)
// Each chunk now has a list of `Records`, which are flattened into one by the `.fold`.
```

It should be fixed in https://github.com/mit-pdos/distributary/pull/37/commits/cc996f285aa5ce57c9e2ed62729295cc18c60080 - which adds a second `flat_map`, and collects each chunk into a `Records` before passing it on to the packet.

dataflow/src/domain/mod.rs
ekmartin: My reasoning here was to avoid crashing when `.recovery()` is called the first time a Soup application starts up (since there are no logs at that point). A warning might be better, but that would mean outputting that when the application starts up the very first time as well, which might be slightly confusing.

ekmartin:
@ms705: That sounds about right! I left a comment regarding 1 [here](https://github.com/mit-pdos/distributary/pull/37#discussion_r149761105).

> While handling a recovery message, a domain does not process normal writes. However, if writes arrived to the domain before the StartRecovery control packet arrived, there might be some writes stored already.

Yep, that's how I understand it as well. Since `handle_recovery` runs from start to finish synchronously, that should block the polling loop in `Domain.run()` and ensure that no other packets get processed. 

Making sure that no other packets can arrive before recovery starts altogether sounds like a good future improvement, perhaps by moving from an explicit `.recover()` call to a parameter on the controller?

dataflow/src/domain/mod.rs
ms705: I'm happy to accept a warning ("No logs for node {} found, starting out empty") on first run, I think!

After all, the common case for startup of a useful database is that it already has data in it ;-)

ekmartin: Indeed! Added a warning in 9bf5859cd925cf6978e7498cae2b15824d9ebbb3.

ms705:
@ekmartin I think we're happy to merge this now :+1: I'll fix the new conflicts later tonight (and merge) unless you get there first.

ekmartin:
@ms705: rebased 👍

ekmartin:
The recovery tests started failing randomly at [this assertion](https://github.com/mit-pdos/distributary/blob/0eb5ba552f50320bcaebb7e7ec329045dc411bd2/dataflow/src/ops/project.rs#L157) after rebasing (repeated the tests successfully 100 times pre-rebase to compare) - I'll investigate.

src/tests.rs
```diff
@@ -53,6 +54,12 @@ fn get_log_name(base: &str) -> String {
     format!("{}-{}-{}", base, current_time.sec, current_time.nsec)
 }
 
+fn delete_log_files(log_name: String) {
+    for log_path in glob::glob(&format!("./{}-*.json", log_name)).unwrap() {
```
jonhoo: I feel like this is going to come back to bite us.

ekmartin: I'm still trying to figure out the test bug I mentioned above, but I realized that I probably wasn't actually taking down the graph before running `.recover()` in the tests. Now that I do `DeleteOnExit` naturally won't work, since that means the logs are deleted when the first graph is dropped.

I'm currently prefixing the logs with each tests' name, and added a timestamp in https://github.com/mit-pdos/distributary/pull/37/commits/458f854bf3542ae6cb101b95014b7454c7935ff0 to ensure that panics wouldn't break future test runs (since the test files wouldn't get cleaned up). 

Why do you feel like this might cause issues in the future? Definitely open to improvement suggestions!

jonhoo: I just get nervous when I see anything that does `rm` and contains a `*`. I'm sure it's fine, especially given that you both prefix and suffix the glob.

ms705: Where does `delete_log_files` get called from in the new world order? Is it just important for tests (where timestamps in the file names would avoid clashes, even though the developer would have to manually clean the logs), or does it serve a more general purpose?

ms705:
Merging #43 should have fixed the failures that you saw.
Once we've sorted out the log file deletion issue to our satisfactions, I think we should merge this.

ekmartin:
@ms705 and @jonhoo: I added a [commit](https://github.com/mit-pdos/distributary/pull/37/commits/03774c5048b5f6e310cff897690b3461bf6feff4) that implements the log file clean-up in tests using `Drop` instead. This should mimic the behavior of calling `delete_log_files` manually, but should ensure that it happens in the case of assertion failures and so on as well.

ms705:
:+1: @jonhoo?

dataflow/src/checktable/mod.rs
```diff
@@ -285,6 +285,27 @@ impl CheckTable {
         }
     }
 
+    pub fn recover(&mut self, base: NodeIndex) -> (i64, Option<Box<HashMap<domain::Index, i64>>>) {
```
jonhoo: I'd like a comment over this function explaining exactly what it does.
I *think* what it does is reserve a single timestamp for the recovery, and then make it the latest timestamp for *every* base, essentially to bootstrap the timestamping logic. However, I don't know why it is *necessary*? Aren't the bases all synchronized at the beginning with `ts=0`? When the system crashes and comes back up, there are no writes anywhere in the graph, so taking a time seems unnecessary?

Cargo.toml
```diff
@@ -78,6 +78,8 @@ default-features = false
 backtrace = { version = "0.3.2", features = ["serialize-serde"] }
 toml = "0.4.1"
 diff = "0.1.10"
+time = "0.1.38"
```
jonhoo: This seems like an unfortunate extra dependency. Still reading, so don't know yet why it's there. Maybe there's a good reason. But if there isn't, we should try to remove it.

dataflow/src/domain/mod.rs
```diff
@@ -30,6 +34,7 @@ pub struct Config {
 }
 
 const BATCH_SIZE: usize = 256;
+const RECOVERY_BATCH_SIZE: usize = 512;
```
jonhoo: Why not larger? We know nothing else is happening in the graph until it has recovered anyway.

dataflow/src/domain/mod.rs
```diff
@@ -1566,6 +1574,80 @@ impl Domain {
         }
     }
 
+    fn handle_recovery(&mut self) {
+        let node_info: Vec<_> = self.nodes
+            .iter()
+            .map(|(index, node)| {
+                let n = node.borrow();
+                (index.clone(), n.global_addr(), n.is_transactional())
```
jonhoo: I don't think you need the `clone` here, because `index` is `Copy`. I think you can just do `*index`, or match against `&index` in the map pattern.

dataflow/src/domain/mod.rs
```diff
@@ -1566,6 +1574,80 @@ impl Domain {
...
+            })
+            .collect();
```
jonhoo: Why collect if you're directly iterating over it below anyway?

dataflow/src/domain/mod.rs
```diff
@@ -1566,6 +1574,80 @@ impl Domain {
...
+
+        for (local_addr, global_addr, is_transactional) in node_info {
+            let checktable = self.transaction_state.get_checktable();
```
jonhoo: Move outside the loop?

dataflow/src/domain/mod.rs
```diff
@@ -1566,6 +1574,80 @@ impl Domain {
...
+
+                    continue;
+                }
+                Err(e) => panic!("Could not open log file {:?}: {}", path, e),
+            };
+
+            BufReader::new(file)
+                .lines()
+                .filter_map(|line| {
+                    let line = line.unwrap();
```
jonhoo: Can you make this an `expect` instead, and then state why we're expecting it to not fail?

jonhoo: Alternatively, given that you're already using `filter_map`, just return `None` if `line.is_err()`. Something like:
```rust
line.ok().and_then(|line| -> Option<Vec<Records>> { serde_json::from_str(&line).ok() })
```

dataflow/src/domain/mod.rs
```diff
@@ -1566,6 +1574,80 @@ impl Domain {
...
+                .flat_map(|r| r)
+                // Merge individual records into batches of RECOVERY_BATCH_SIZE:
+                .chunks(RECOVERY_BATCH_SIZE)
+                .into_iter()
+                // Then create Packet objects from the data:
+                .map(|chunk| {
+                    let data: Records = chunk.collect();
+                    let link = Link::new(local_addr, local_addr);
+                    if is_transactional {
+                        let (ts, prevs) = checktable.recover(global_addr).unwrap();
```
jonhoo: Hmm.. Shouldn't it be possible to just claim a single timestamp for the *entire* recovery (across all base nodes)? @fintelia may be able to shed some light.

dataflow/src/domain/mod.rs
```diff
@@ -1566,6 +1574,80 @@ impl Domain {
...
+                        }
+                    } else {
+                        Packet::Message {
+                            link,
+                            data,
+                            tracer: None,
+                        }
+                    }
+                })
+                .for_each(|packet| self.handle(box packet));
```
jonhoo: I'm pretty sure this could deadlock if you had an A-B-A domain assignment. But we already disallow that elsewhere, so probably not an issue.

examples/basic-recipe.rs
```diff
@@ -29,21 +44,24 @@ fn main() {
...
+        let title = "test title";
+        let url = "http://pdos.csail.mit.edu";
+        article
+            .put(vec![aid.into(), title.into(), url.into()])
+            .unwrap();
+    }
+
+    // Then create a new vote:
+    println!("Casting vote...");
+    let uid = time::get_time().sec;
```
jonhoo: Why is the `time` crate needed for this? Why is `SystemTime::now().duration_since(UNIX_EPOCH).as_secs()` not sufficient using `std::time`?

src/controller/mod.rs
```diff
@@ -518,6 +519,23 @@ impl ControllerInner {
...
+            let node = &self.ingredients[*index];
+            let domain = self.domains.get_mut(&node.domain()).unwrap();
+            domain.send(box payload::Packet::StartRecovery).unwrap();
+        }
+
+        for (_name, index) in self.inputs(()).iter() {
+            let node = &self.ingredients[*index];
+            let domain = self.domains.get_mut(&node.domain()).unwrap();
+            domain.wait_for_ack().unwrap();
+        }
```
jonhoo: I'm pretty sure this can deadlock. Imagine two domains that both have egresses to the other. Both are told to recover at the same time. If there's enough data, they will eventually fill up the channels between one another, and deadlock.

I think you'll have to wait for an ack from each node recovery sequentially unfortunately. At least until we can come up with a cleverer scheme.

src/tests.rs
```diff
@@ -27,22 +30,51 @@ use controller::recipe::{ActivationResult, Recipe};
...
+use std::fs;
 use std::collections::HashMap;
 
 const DEFAULT_SETTLE_TIME_MS: u64 = 100;
 
-fn get_settle_time() -> time::Duration {
+// Suffixes the given log prefix with a timestamp, ensuring that
+// subsequent test runs do not reuse log files in the case of failures.
+fn get_log_name(prefix: &str) -> String {
+    let current_time = time::get_time();
```
jonhoo: See note further up about using `std::time` for this.

src/tests.rs
```diff
@@ -376,6 +409,181 @@ fn it_works_with_arithmetic_aliases() {
 }
 
 #[test]
+fn it_recovers_persisted_logs() {
```
jonhoo: I like this test.

jonhoo:
See inline comments. Overall looks good though!

dataflow/src/checktable/mod.rs
fintelia: It actually reserves a timestamp for the recovery of a specific base, and then marks that time as the latest timestamp for all columns in that base. This is necessary because at `ts=0` all base nodes are empty, and the recovery messages flowing through the graph are implemented as writes (and thus each must have some distinct timestamp assigned to them).

dataflow/src/domain/mod.rs
fintelia: Lifetime issues: `self.handle()` can't by called while `self.nodes` is borrowed

dataflow/src/domain/mod.rs
fintelia: This is a limitation of our transaction logic. Every timestamped message must originate from only a single base node. Further, every message from a base must also have its own distinct timestamp (which is why there can't be one timestamp per base either).

dataflow/src/domain/mod.rs
ekmartin: Fixed! Didn't need the `.clone` at all, like you said.

dataflow/src/domain/mod.rs
ekmartin: Done!

dataflow/src/domain/mod.rs
ekmartin: I changed it to an `.expect` as I thought it might be good to blow up in the case that this ever happens (IO-errors, or invalid UTF-8 in the file - if I understand it correctly).
Let me know if you disagree though.

https://github.com/mit-pdos/distributary/pull/37/files#diff-f3207323d6f34ad3c124fbcb4cb87a40R1612

examples/basic-recipe.rs
ekmartin: Good idea - fixed in https://github.com/mit-pdos/distributary/pull/37/commits/a6185971bdf70c230c4f16d56307cc509a0f30b1.

src/controller/mod.rs
ekmartin: That's unfortunate :/ changed back to sequential recovery in https://github.com/mit-pdos/distributary/pull/37/commits/8126b20de628f3f86d1b863ceca7f30128efb7ed.

dataflow/src/domain/mod.rs
ekmartin: @fintelia suggested 512 - I'm not completely sure what the trade-offs are here. What do you think would be a good number?

src/controller/mod.rs
ms705: I agree that sequential recovery is the safe option. Thinking about it, though, I don't think there is (currently) any way that the described deadlock would occur: it would require that a new domain adds a node into an existing domain 
 *below* its own nodes: i.e., an existing `A -> B` relation turns into either `A -> B -> A` (not permitted) or independent `A -> B`, `B -> A` paths (which require `B` to add an edge, and thus a node, into `A`). I don't think our current assignment logic can ever do this?

ms705:
I think we're ready to merge this, so I will unless I hear any objections :-)

src/controller/mod.rs
jonhoo: I'm hesitant to say that the current logic can't do that actually. It does funky things when there are joins (which I think essentially ends up as placing the join in the same domain as one of its ancestors).

jonhoo:
My only hesitation is https://github.com/mit-pdos/distributary/pull/37#discussion_r150422839

ms705:
That sounds like a broader limitation of the transactional logic, though, so not something we would want to fix within the scope of this PR. I agree that it would be nice to recover at a single timestamp or a single timestamp per base, but I suspect that doing so would involve some invasive changes unrelated to the basic recovery implemented here. Let's merge this first, and then improve it.

fintelia:
@ms705 It is not clear to me that we'd even want recovery to use only a single timestamp. Recall that our buffering logic means that that there can be no pipeline parallelism in the processing of messages for a single timestamp

ms705:
@fintelia Good point! This would also cause the timestamp-per-base scheme to buffer all writes from a single base until another base (with a "prior" timestamp) has finished recovering, which would in effect serialize the recovery at joins.

PR #38, opened 2017-10-22
Use rustfmt-nightly 0.2.9

ekmartin:
This fixes the Travis build. Was released a week ago, so slightly newer than the hash that was being used before.


ekmartin:
I changed this to use rev `9754bcb535fa84c000fe78cb91b30c3f661fc849` instead, since that's what's being [used in master](https://github.com/mit-pdos/distributary/commit/2e4c0d0857b4ca754b978334e46984523d4b7d2d) now.

PR #39, opened 2017-10-22
Add support for aliases in arithmetic expressions

ekmartin:
Not sure if this is the right place to solve this, but it basically sets the column name of the output column to the alias if it exists - or `arithmetic` if it doesn't (I don't think the actual `arithmetic` name is being used for anything?).

This is sort of similar to what's done in [SqlToMirConverter.make_grouped_node](https://github.com/mit-pdos/distributary/blob/master/src/sql/mir.rs#L585-L595).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -126,7 +126,7 @@ dependencies = [
  "mio 0.6.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "mysql 11.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "net2 0.2.31 (registry+https://github.com/rust-lang/crates.io-index)",
- "nom_sql 0.0.1 (git+https://github.com/ms705/nom-sql.git?rev=13744adf972e42741b5e7146db46e8ed7353a15d)",
+ "nom_sql 0.0.1 (git+https://github.com/ms705/nom-sql.git?rev=fcf7faa4bbb5592b0c1a4991278c0a6cb62926d4)",
  "rand 0.3.17 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "slog 2.0.12 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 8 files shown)


ms705:
Apologies, I missed this! Looks good, though it will need a rebase onto changed `master`. Given he small change, it should rebase cleanly (fingers crossed).

Do you want to go ahead and do this so we can merge?

ms705:
P.S.: I guess the "correct" solution to the no-alias case for both this and function columns is to use the actual column definition expression as the name (i.e., `price / 100`); certainly that's what MySQL and friends do. With a `to_string()` implementation on `ArithmeticExpression`, this should be doable.

Cargo.lock
```diff
@@ -1,14 +1,3 @@
-[root]
```
ekmartin: Not sure if it matters, but running `cargo build` on the latest nightly (11-07) seems to shuffle things around a bit in the lockfile.

ekmartin:
@ms705: no worries at all - I've rebased it now. I can send a separate PR for the name change. Just to make sure I'm understanding this correctly though, the names for e.g. arithmetic isn't actually shown anywhere right now right? Should literals get their column name changed to `value.to_string()` as well?

Cargo.lock
jonhoo: I'll push a commit to `master` that makes this change so we avoid cluttering the PR

jonhoo: Done in 3188e79991563a0e6fa35fb96c05986303ee48d1, so if you rebase you should get it.

ekmartin: Thanks, done!

ms705:
@ekmartin That's correct -- the current status is that non-aliased arithmetic and literal columns have string constants as names. In practice, real databases typically use a best-effort approach to generate unique column names in the absence of aliases -- consider, e.g., in MySQL:
```
mysql> SELECT 42, 42 * 3;
+----+--------+
| 42 | 42 * 3 |
+----+--------+
| 42 |    126 |
+----+--------+

mysql> SELECT REPEAT('a',1) UNION SELECT REPEAT('b',10);
+---------------+
| REPEAT('a',1) |
+---------------+
| a             |
| bbbbbbbbbb    |
+---------------+
```
It's easy for us to do the same for arithmetic and literal expressions; we already do for function expressions.

ms705:
Also, totally fine to merge this change and do the name fixup in a different one :+1:

PR #41, opened 2017-11-08
preserve migration state

larat7:
Avoid doing some computations over all nodes, instead only do them over only _new_ nodes.
Also, do `O(#bases)` passes over the graph instead of `O(#nodes)`.


src/flow/migrate/transactions.rs
```diff
@@ -36,12 +34,6 @@ fn count_base_ingress(
                 })
                 .sum();
 
-            // Domains containing a base node will get a single copy of each packet sent to it.
-            if graph[base].domain() == *domain {
-                assert_eq!(num_paths, 0);
-                num_paths = 1;
-            }
```
jonhoo: This seems like a semantic change, no? Why is it okay to remove it?

src/flow/migrate/transactions.rs
```diff
@@ -89,21 +80,51 @@ fn has_path(graph: &Graph, source: NodeIndex) -> HashSet<(NodeIndex, NodeIndex)>
...
+    old: &mut HashMap<DomainIndex, (IngressFromBase, EgressForBase)>,
+    new: HashMap<DomainIndex, (IngressFromBase, EgressForBase)>
+) {
+    for (di, (new_ingress, new_egress)) in new {
+        let entry = old.entry(di).or_insert((HashMap::new(), HashMap::new()));
+        let old_ingress = &mut entry.0;
+        let old_egress = &mut entry.1;
+
+        for (base, v) in new_egress {
+            let e = old_egress.entry(base).or_insert(vec![]);
```
jonhoo: You want `or_insert_with(Vec::new)` here. Or `or_default()`.

src/flow/migrate/transactions.rs
```diff
@@ -89,21 +80,51 @@ fn has_path(graph: &Graph, source: NodeIndex) -> HashSet<(NodeIndex, NodeIndex)>
...
+            (*e).extend(v);
+        }
+
+        for (base, v) in new_ingress {
+            let e = old_ingress.entry(base).or_insert(0);
+            *e += v;
+
+            // Domains containing a base node will get a single copy of each packet sent to it.
+            if graph[base].domain() == di {
+                *e = 1;
```
jonhoo: Ah, the check moved here.

src/flow/migrate/transactions.rs
jonhoo: nvm -- found the moved check below.

larat7: It was just moved, since it doesn't actually compute anything over the ingress nodes.
See https://github.com/mit-pdos/distributary/pull/41/files/56f572e445bf2517185fb689dbf4f37c75232f00#diff-5af68982bf87be5004d3320a768608a8R122

src/flow/migrate/transactions.rs
```diff
@@ -89,21 +80,51 @@ fn has_path(graph: &Graph, source: NodeIndex) -> HashSet<(NodeIndex, NodeIndex)>
...
+    new: HashMap<DomainIndex, (IngressFromBase, EgressForBase)>
+) {
+    for (di, (new_ingress, new_egress)) in new {
+        let entry = old.entry(di).or_insert((HashMap::new(), HashMap::new()));
+        let old_ingress = &mut entry.0;
+        let old_egress = &mut entry.1;
+
+        for (base, v) in new_egress {
+            let e = old_egress.entry(base).or_insert(vec![]);
+            (*e).extend(v);
```
jonhoo: I don't think the `(*` should be necessary here?

src/flow/migrate/transactions.rs
```diff
@@ -89,21 +80,51 @@ fn has_path(graph: &Graph, source: NodeIndex) -> HashSet<(NodeIndex, NodeIndex)>
 pub fn analyze_graph(
     graph: &Graph,
     source: NodeIndex,
-    domain_nodes: HashMap<DomainIndex, Vec<(NodeIndex, bool)>>,
-) -> HashMap<DomainIndex, (IngressFromBase, EgressForBase)> {
+    domain_nodes: HashMap<DomainIndex, Vec<NodeIndex>>,
```
jonhoo: I'd rename this to `domain_new_nodes`.

src/flow/mod.rs
```diff
@@ -1091,63 +1100,54 @@ impl<'a> Migration<'a> {
...
             });
 
         // Assign local addresses to all new nodes, and initialize them
-        let mut local_remap = HashMap::new();
-        let mut remap = HashMap::new();
-        for (domain, nodes) in &mut domain_nodes {
+        for (domain, nodes) in &mut domain_new_nodes {
             // Number of pre-existing nodes
-            let mut nnodes = nodes.iter().filter(|&&(_, new)| !new).count();
+            let mut nnodes = mainline.remap.entry(*domain).or_insert_with(HashMap::new).len();
```
jonhoo:
```rust
mainline.remap.get(domain).map(|n| n.len()).unwrap_or(0);
```

src/flow/mod.rs
```diff
@@ -1091,63 +1100,54 @@ impl<'a> Migration<'a> {
...
 
-            if nnodes == nodes.len() {
+            if nodes.is_empty() {
```
jonhoo: Hmm.. This is not the same check any more. Why did you change it?

src/flow/mod.rs
```diff
@@ -1191,14 +1191,28 @@ impl<'a> Migration<'a> {
...
         // println!("{}", mainline);
 
-        let mut uninformed_domain_nodes = domain_nodes.clone();
-        let deps = migrate::transactions::analyze_graph(
+        migrate::transactions::analyze_graph(
             &mainline.ingredients,
             mainline.source,
-            domain_nodes,
+            domain_new_nodes,
+            &mut mainline.deps,
```
jonhoo: Any particular reason you chose to do the merge inside `analyze_graph` instead of returning `new` from there and doing the merge here (which would save you an argument to `analyze_graph`)?

src/flow/mod.rs
```diff
@@ -1191,14 +1191,28 @@ impl<'a> Migration<'a> {
...
+            .node_indices()
+            .filter(|&ni| ni != mainline.source)
+            .filter(|&ni| !mainline.ingredients[ni].is_dropped())
+            .map(|ni| {
+                (mainline.ingredients[ni].domain(), ni, new.contains(&ni))
+            })
+            .fold(HashMap::new(), |mut dns, (d, ni, new)| {
+                dns.entry(d).or_insert_with(Vec::new).push((ni, new));
+                dns
+        });
```
jonhoo: This seems a little unfortunate. I wonder if we could rewrite this in terms of `new` too.

larat7: This is used by `inform` which I think needs to inform all domains and do some computation over new nodes. So, it might be able to rewrite this, but the next bottle neck in the system is actually `materialization::commit`, so I don't think we should worry to much about it.

src/flow/mod.rs
larat7: Not particularly, although now that I think about it, I should probably rename it to `update_deps` or something.

jonhoo: I'd say make it `analyze_changes`, have it return `new`, and update in caller. Improves modularity.

src/flow/mod.rs
jonhoo: Do you know what in `commit` is slow? I know it does a topological traversal of the entire graph, and that's bound to come back to bite us.

jonhoo: As will the countless calls to `any_partial`.

larat7: I haven't profiled it in depth. With 5k universes, a single call takes ~60ms. I don't think the traversal of the graph is that bad (if we only do it once), since `has_path` does `#bases` traversals and only takes ~20ms. Not sure about `any_partial` though.

PR #43, opened 2017-11-10
Sort new domains before creating nodes

ekmartin:
This ensures deterministic indices and so on.


jonhoo:
Just change
```rust
let mut sorted_new = new.iter().collect::<Vec<_>>();
```
to
```rust
let mut sorted_new = new.iter().cloned().collect::<Vec<_>>();
```

PR #45, opened 2017-11-17
Use appropriate names for literals and arithmetic expressions

ekmartin:
This uses `.to_string()` on nom-sql's literals and arithmetic expressions for column names. Added a rustfmt commit too for Travis, but I'll remove it if someone runs it on master before this is merged.

### DIFF
--- a/src/controller/sql/mod.rs
+++ b/src/controller/sql/mod.rs
@@ -618,6 +618,8 @@ impl SqlIncorporator {
             // other kinds of queries *do* require their referred tables to exist!
             ref q @ SqlQuery::CompoundSelect(_) |
             ref q @ SqlQuery::Select(_) |
+            ref q @ SqlQuery::Update(_) |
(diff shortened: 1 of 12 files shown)


src/controller/sql/mod.rs
```diff
@@ -618,6 +618,8 @@ impl SqlIncorporator {
             // other kinds of queries *do* require their referred tables to exist!
             ref q @ SqlQuery::CompoundSelect(_) |
             ref q @ SqlQuery::Select(_) |
+            ref q @ SqlQuery::Update(_) |
```
ekmartin: Let me know how you want to handle the new `Update` and `Delete` expressions. I added `unreachable!` in a couple of other places, but included them here since it seemed to make sense.

ms705: Yes, this is the right way. `unreachable!` would be fine here too, but this is correct if we ever choose to support them later.

PR #46, opened 2017-11-23
Enable local workers in the example

ekmartin:
Also adds back logging - not sure if this was removed intentionally, but seems useful to me at least?

### DIFF
--- a/examples/basic-recipe.rs
+++ b/examples/basic-recipe.rs
@@ -28,6 +28,8 @@ fn main() {
 
     // set up Soup via recipe
     let mut builder = ControllerBuilder::default();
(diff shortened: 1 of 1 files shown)


ms705:
Looks good to me! :+1: 

OOI, what happens if `set_local_workers()` isn't called? The controller presumably waits for remote workers to connect?

ekmartin:
At the moment it panics, since `nworkers` is 0. I tried setting `nworkers` to 2 first, but I realized that nothing starts workers by itself, which means it waits for remote workers like you said.

jonhoo:
That `assert` should probably be given a description saying that there have to either be local or remote workers.

PR #47, opened 2017-11-23
Re-enable unused warnings, and fix existing instances of them

fintelia:
A lot of these changes are in SQL/MIR which I'm not too familiar with, so I didn't want to push to master without confirmation. For the most part I just tagged functions/structs with `#[allow(unused)]` but it may be the case that many of them should be remove instead.

### DIFF
--- a/src/controller/mod.rs
+++ b/src/controller/mod.rs
@@ -1,20 +1,20 @@
-use channel::poll::{PollEvent, PollingLoop, ProcessResult, RpcPollEvent, RpcPollingLoop};
+use channel::poll::{PollEvent, PollingLoop, ProcessResult, RpcPollingLoop};
 use channel::tcp::TcpSender;
 use channel;
 use dataflow::prelude::*;
-use dataflow::{backlog, checktable, node, payload, DomainConfig, PersistenceParameters, Readers};
+use dataflow::{checktable, node, payload, DomainConfig, PersistenceParameters, Readers};
 use dataflow::ops::base::Base;
 use dataflow::statistics::GraphStats;
 use worker::Worker;
 
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::error::Error;
-use std::io::{Read, Write};
+use std::io::Read;
 use std::net::{IpAddr, SocketAddr};
 use std::time::{Duration, Instant};
-use std::sync::{Arc, Barrier, Mutex};
-use std::sync::mpsc::{self, Receiver, Sender};
-use std::{fmt, io, thread, time};
+use std::sync::{Arc, Mutex};
+use std::sync::mpsc::{self, Sender};
+use std::{io, thread, time};
 
 use futures::{Future, Stream};
 use hyper::Client;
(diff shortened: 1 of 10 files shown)


src/controller/recipe/mod.rs
```diff
@@ -62,6 +62,7 @@ fn hash_query(q: &SqlQuery) -> QueryID {
     h.finish()
 }
 
+#[allow(unused)]
```
ms705: I'm not sure how this is unused -- `self.recipe` on `ControllerInner` stores a `Recipe`, so it should use this type?

src/controller/sql/mod.rs
```diff
@@ -340,7 +344,7 @@ impl SqlIncorporator {
 
         // TODO(malte): we currently need to remember these for local state, but should figure out
         // a better plan (see below)
-        let fields = mir.leaf
+        let _fields = mir.leaf
```
ms705: This can indeed be removed, as `register_query` now contains this code.

src/controller/sql/mod.rs
```diff
@@ -557,7 +561,7 @@ impl SqlIncorporator {
         &mut self,
         q: SqlQuery,
         query_name: String,
-        is_leaf: bool,
+        _is_leaf: bool,
```
ms705: Good catch, this is a bug. The call to `self.add_select_query()` on line 653 should use `is_leaf` instead of always passing `true`.

ms705:
Looks good! :+1: I think we should retain all currently unused code in `sql` and `mir` -- it either is actually used (as with `Recipe`), is unused because of a bug (`is_leaf` argument to `nodes_for_named_query`), or should be hooked back into higher layers (as with the reuse-related methods on `SqlIncorporator`).

I'll fix the actual instance of unused code, and the `is_leaf` bug in a separate change, so we can merge this one.

ms705:
(OOI, this still seems to fail on Travis due to many unused imports, and due to rustfmt issues. I guess there are more places that need fixing up?)

src/controller/mod.rs
```diff
@@ -508,50 +515,6 @@ impl ControllerInner {
...
-    ///     deleted once the `Controller` is dropped. Useful for tests.
-    ///  3. `DurabilityMode::MemoryOnly`: no writes to disk, store all writes in memory.
-    ///     Useful for baseline numbers.
-    ///
-    /// `queue_capacity` indicates the number of packets that should be buffered until
-    /// flushing, and `flush_timeout` indicates the length of time to wait before flushing
-    /// anyway.
-    ///
-    /// Must be called before any domains have been created.
-    pub fn with_persistence_options(&mut self, params: PersistenceParameters) {
```
ms705: This does not appear to be unused: tests fail to compile with it removed.

ms705: (I believe it is never used *outside* tests currently, and hence we get a warning when compiling the library.)

src/controller/mod.rs
```diff
@@ -515,6 +515,34 @@ impl ControllerInner {
...
+    ///     deleted once the `Controller` is dropped. Useful for tests.
+    ///  3. `DurabilityMode::MemoryOnly`: no writes to disk, store all writes in memory.
+    ///     Useful for baseline numbers.
+    ///
+    /// `queue_capacity` indicates the number of packets that should be buffered until
+    /// flushing, and `flush_timeout` indicates the length of time to wait before flushing
+    /// anyway.
+    ///
+    /// Must be called before any domains have been created.
+    pub fn with_persistence_options(&mut self, params: PersistenceParameters) {
```
jonhoo: If these are indeed only intended to be used by tests, just prefix them with `#[cfg(test)]`

PR #48, opened 2017-11-23
enable partial through unions

larat7:


larat7:
@jonhoo was this what you had in mind?

jonhoo:
Yes! Looks good, though I think you can now remove the `if let` that has the comment:
```rust
// request came in from reader -- forward
```
And instead always do the call that's in the `else`.
Do the tests all pass locally for you?

PR #49, opened 2017-11-30
Implement From<usize> for DataType

ekmartin:
Just thought it was annoying having to do `(i as i32).into()` in tests etc. - feel free to close if there's a good reason why it isn't implemented though!


jonhoo:
I think the reason was principally because we might want to support unsigned numbers as its own datatype at some point. But it might not be important, since in those cases you *would* want a `usize` to be turned into the appropriate type.

PR #50, opened 2017-12-04
Explicitly tag public queries in recipes

ms705:
This changes the recipe syntax to allow two forms of named query:

 1. `QUERY q_name: SELECT ...`
 2. `q_name: SELECT ...`

The first is a *public* query, i.e., one that a client wishes to read from and for which we must maintain a leaf materialization. The second is a *shorthand*, nested query that clients won't query, but which other queries might refer to (treating it as a view/table). Only public queries receive readable leaf materializations (= reader nodes in the data flow).

Note that this is a breaking change that breaks all existing recipes. I've fixed the unit tests, the example, and "vote"; other branches need to take care when merging (@larat7 @fintelia @jonhoo @ekmartin)


ms705:
(Turning this into a PR so that everyone sees the syntax change.)

PR #51, opened 2017-12-04
Materialization status in graphviz

ekmartin:
The snapshotting implementation needs to know which domains have at least one materialized node, but I thought it might be useful to show it in Blender's graphviz output as well.

I'm not sure if the way I'm retrieving materialization status for a node is correct, so let me know if there's a better way to achieve the same. 

This is what it outputs for the example:
image

Feel free to bikeshed on how it should be shown in the output - currently it uses a half circle  for partial and a full circle for full materializations.

### DIFF
--- a/dataflow/src/node/debug.rs
+++ b/dataflow/src/node/debug.rs
@@ -17,7 +17,11 @@ impl fmt::Debug for Node {
 }
 
 impl Node {
-    pub fn describe(&self, idx: NodeIndex) -> String {
+    pub fn describe(
+        &self,
+        idx: NodeIndex,
+        materialization_status: Option<MaterializationStatus>,
+    ) -> String {
         let mut s = String::new();
         let border = match self.sharded_by {
             Sharding::ByColumn(_, _) | Sharding::Random(_) => "filled,dashed",
@@ -32,6 +36,12 @@ impl Node {
                 .unwrap_or("white".into())
         ));
 
+        let materialized = match materialization_status {
+            None => "",
+            Some(MaterializationStatus::Partial) => "| ◑",
+            Some(MaterializationStatus::Full) => "| ⬤",
+        };
+
         let addr = match self.index {
             Some(ref idx) => if idx.has_local() {
                 format!("{} / {}", idx.as_global().index(), **idx)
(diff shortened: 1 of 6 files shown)


ms705:
NICE! I'm a fan :+1:

dataflow/src/node/debug.rs
```diff
@@ -40,33 +50,42 @@ impl Node {
...
-            NodeType::Egress { .. } => s.push_str(&format!("{{ {} | (egress) }}", addr)),
+            NodeType::Ingress => s.push_str(&format!(
+                "{{ {{ {} {} }} | (ingress) }}",
+                addr,
+                materialized
+            )),
+            NodeType::Egress { .. } => {
+                s.push_str(&format!("{{ {{ {} {} }} | (egress) }}", addr, materialized))
+            }
             NodeType::Sharder { .. } => s.push_str(&format!("{{ {} | (sharder) }}", addr)),
```
ekmartin: Can a `Sharder` be materialized?

jonhoo: No, a `Sharder` cannot be materialized afaik.

dataflow/src/node/debug.rs
```diff
@@ -40,33 +50,42 @@ impl Node {
...
             NodeType::Source => s.push_str("(source)"),
             NodeType::Dropped => s.push_str("✗"),
-            NodeType::Ingress => s.push_str(&format!("{{ {} | (ingress) }}", addr)),
-            NodeType::Egress { .. } => s.push_str(&format!("{{ {} | (egress) }}", addr)),
+            NodeType::Ingress => s.push_str(&format!(
+                "{{ {{ {} {} }} | (ingress) }}",
+                addr,
+                materialized
+            )),
+            NodeType::Egress { .. } => {
```
ms705: I don't think an `Egress` node would currently ever be materialized (@jonhoo -- ?), though this code is obviously still correct in that case.

src/controller/migrate/materialization/mod.rs
```diff
@@ -404,6 +404,21 @@ impl Materializations {
         assert!(replay_obligations.is_empty());
     }
 
+    /// Retrieves the materialization status of a given node, or None
+    /// if the node isn't materialized.
+    pub fn get_status(&self, index: &NodeIndex, node: &Node) -> MaterializationStatus {
+        let is_materialized = self.have.contains_key(index)
+            || node.with_reader(|r| r.is_materialized()).unwrap_or(false);
```
ms705: Correct, but aren't readers always materialized? (`is_materialized`returns a bool, so it doesn't differentiate full/partial either)

dataflow/src/node/debug.rs
jonhoo: Yeah, `Egress` is never materialized.

dataflow/src/node/debug.rs
```diff
@@ -40,10 +50,15 @@ impl Node {
...
         match self.inner {
             NodeType::Source => s.push_str("(source)"),
             NodeType::Dropped => s.push_str("✗"),
-            NodeType::Ingress => s.push_str(&format!("{{ {} | (ingress) }}", addr)),
+            NodeType::Ingress => s.push_str(&format!(
+                "{{ {{ {} {} }} | (ingress) }}",
+                addr,
+                materialized
+            )),
             NodeType::Egress { .. } => s.push_str(&format!("{{ {} | (egress) }}", addr)),
```
ekmartin: Removed `materialization` from `Egress`

ms705:
LGTM :+1: Let's merge?

PR #52, opened 2017-12-05
Zookeeper

jonhoo:


benchmarks/vote/graph.rs
```diff
@@ -64,8 +65,8 @@ pub fn make(s: Setup, persistence_params: PersistenceParameters) -> Graph {
         g.disable_sharding();
     }
     g.set_persistence(persistence_params);
-    g.set_nworkers(s.nworkers);
-    let graph = g.build();
+    // g.set_nworkers(s.nworkers);
```
jonhoo: `set_nworkers` needs to be supported.

benchmarks/vote/vote.rs
```diff
@@ -288,11 +288,11 @@ fn main() {
 
     // prepare putters
     let putters: Vec<_> = {
-        let mix_getters = (0..new_vote_receivers.len())
-            .map(|_| mix.as_ref().map(|_| g.graph.get_getter(g.end).unwrap()));
+        let mix_getters: Vec<_> = (0..new_vote_receivers.len())
+            .map(|_| mix.as_ref().map(|_| g.graph.get_getter(g.end).unwrap())).collect();
```
jonhoo: It'd be nice if we can avoid this collect.

consensus/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+mod local;
+pub use zk::ZookeeperAuthority;
+pub use local::LocalAuthority;
+
+const CONTROLLER_KEY: &'static str = "/controller";
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct Epoch(i64);
+
+pub trait Authority: Send + Sync {
```
jonhoo: Document which methods should be called from where.

consensus/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// Attempt to become leader. Return leader epoch if successful, None otherwise.
+    fn become_leader(&self, payload_data: Vec<u8>) -> Option<Epoch>;
```
jonhoo: Should block if failed to become leader.

consensus/src/local.rs
```diff
@@ -0,0 +1,115 @@
...
+        while !inner.keys.contains_key(CONTROLLER_KEY) {
+            inner = self.cv.wait(inner).unwrap();
+        }
+        (ONLY_EPOCH, inner.keys.get(CONTROLLER_KEY).cloned().unwrap())
+    }
+
+    fn await_new_epoch(&self, _: Epoch) -> Epoch {
+        loop {
+            thread::sleep(Duration::from_secs(1000));
+        }
```
jonhoo: This should probably just use `thread::park`

consensus/src/zk.rs
```diff
@@ -0,0 +1,163 @@
...
+impl Watcher for EventWatcher {
+    fn handle(&self, e: WatchedEvent) {
+        if e.keeper_state != KeeperState::SyncConnected {
+            eprintln!("Lost connection to ZooKeeper! Aborting");
+            process::abort();
+        }
+    }
+}
+
+struct ChannelWatcher(Sender<()>);
```
jonhoo: This should be a `Barrier`, or maybe even just `thread::park` + `unpark`

consensus/src/zk.rs
```diff
@@ -0,0 +1,163 @@
...
+            payload_data,
+            Acl::open_unsafe().clone(),
+            CreateMode::Ephemeral,
+        );
+        if let Ok(path) = create {
+            if let Ok(Some(stat)) = self.zk.exists(&path, false) {
+                return Some(Epoch(stat.czxid));
+            }
+        }
+        return None;
```
jonhoo:
```rust
create.and_then(|path| self.zk.exists(&path, false)).ok().map(|stat| Epoch(stat.czxid))
```

consensus/src/zk.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    fn await_new_epoch(&self, current_epoch: Epoch) -> Epoch {
+        loop {
+            let (watcher, rx) = ChannelWatcher::new();
+            match self.zk.exists_w(CONTROLLER_KEY, watcher) {
+                Ok(ref stat) if stat.czxid > current_epoch.0 => return Epoch(stat.czxid),
+                Ok(_) | Err(ZkError::NoNode) => {
+                    let _ = rx.recv();
+                }
+                Err(e) => panic!("{}", e),
```
jonhoo: Seems sad.

consensus/src/zk.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::sync::Arc;
+
+    #[test]
```
jonhoo: `#[allow_fail]`

dataflow/src/checktable/service.rs
```diff
@@ -27,6 +27,28 @@ pub struct TimestampReply {
...
+pub enum ChecktableReply {
+    ApplyBatch(Option<TimestampReply>),
+    Recover(i64, Option<Box<HashMap<domain::Index, i64>>>),
+    ClaimReplayTimestamp(i64, Option<Box<HashMap<domain::Index, i64>>>),
+    Track,
+    PerformMigration(i64, i64, Option<Box<HashMap<domain::Index, i64>>>),
+    AddReplayPaths,
+    ValidateToken(bool),
+}
+
```
jonhoo: Unrelated change?

consensus/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// Returns the epoch and payload data for the current leader, blocking if there is not
+    /// currently as leader.
+    fn get_leader(&self) -> (Epoch, Vec<u8>);
+
+    /// Wait until it is no longer the epoch indicated in `current_epoch`, and then return the new
+    /// epoch.
+    fn await_new_epoch(&self, current_epoch: Epoch) -> Epoch;
+
+    /// Do a non-blocking read at the indicated path.
+    fn try_read(&self, path: &str) -> Option<Vec<u8>>;
```
jonhoo: `path` is a confusing name for this parameter. Perhaps `key`.

src/controller/main.rs
```diff
@@ -0,0 +1,45 @@
...
+    let zookeeper_addr = matches.value_of("zookeeper").unwrap();
+
+    let authority = ZookeeperAuthority::new(&zookeeper_addr);
+    let mut builder = ControllerBuilder::default();
+    builder.set_listen_addr(listen_addr);
+    let _handle = builder.build(authority);
+
+    loop {
+        thread::sleep(Duration::from_secs(5));
+    }
```
jonhoo: `handle.wait()`?

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
+        if let Some((sender, join_handle)) = self.local.take() {
+            let _ = sender.send(ControlEvent::Shutdown);
+            join_handle.join();
+        }
+    }
+}
 
-    /// Set the `Logger` to use for internal log messages.
-    pub fn log_with(&mut self, _: slog::Logger) {}
+struct ServingThread {
```
jonhoo: I'd like documentation on this.

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
+            healthcheck_every: Duration::from_secs(10),
+        }
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+struct ControllerState {
+    config: ControllerConfig,
+    epoch: Epoch,
+    recipe: String,
```
jonhoo: `: ()` + `// TODO`

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
+        );
+
+        let descriptor = ControllerDescriptor {
+            external_addr: external.addr,
+            internal_addr: internal.addr,
+            checktable_addr: checktable,
+        };
+        let campaign = Self::campaign(event_tx.clone(), authority.clone(), descriptor, config);
+
+        let event_tx2 = event_tx.clone();
```
jonhoo: Unnecessary.

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
+        let (tx, rx) = mpsc::channel();
```
jonhoo: Barrier.

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
+            type Error = hyper::Error;
+            type Future = Box<Future<Item = Self::Response, Error = Self::Error>>;
+
+            fn call(&self, req: Request) -> Self::Future {
+                let mut res = Response::new();
+                match (req.method().clone(), req.path().to_owned().as_ref()) {
+                    (Method::Get, "/graph.html") => {
+                        res.headers_mut().set(ContentType::html());
+                        res.set_body(include_str!("graph.html"));
+                        Box::new(futures::future::ok(res))
```
jonhoo: @jonhoo No boxes.

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
+        let builder = thread::Builder::new().name("srv-zk".to_owned());
+        builder
+            .spawn(move || {
+                loop {
+                    // become leader
+                    let current_epoch = match authority.become_leader(descriptor.clone()) {
+                        Some(epoch) => epoch,
+                        None => continue,
+                    };
+                    let initial_state = ControllerState {
```
jonhoo: Move into `None` match case inside the closure.

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use consensus::ZookeeperAuthority;
 
     // Controller without any domains gets dropped once it leaves the scope.
     #[test]
```
jonhoo: `#[allow_fail]`

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
+        let authority = ZookeeperAuthority::new("127.0.0.1:2181/it_works_default");
+        {
+            let c = ControllerBuilder::default().build(authority);
+            thread::sleep(Duration::from_secs(1));
+        }
+        thread::sleep(Duration::from_secs(1));
     }
 
     // Controller with a few domains drops them once it leaves the scope.
     #[test]
```
jonhoo: `#[allow_fail]`

src/controller/mod.rs
```diff
@@ -1441,124 +1321,509 @@ impl Drop for ControllerInner {
...
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use consensus::ZookeeperAuthority;
 
     // Controller without any domains gets dropped once it leaves the scope.
     #[test]
     fn it_works_default() {
```
jonhoo: Repeat these tests for `LocalAuthority`

consensus/src/zk.rs
```diff
@@ -74,7 +74,7 @@ impl Authority for ZookeeperAuthority {
                 .exists_w(CONTROLLER_KEY, UnparkWatcher::new())
                 .is_ok()
             {
-                thread::park();
+                thread::park_timeout(Duration::from_secs(60));
```
jonhoo: You may want this timeout to be shorter. If this race happens, we don't really want to wait a minute for things to recover

fintelia: We're balancing how quickly we respond to a *really* unlikely race with the frequency at which we needlessly poll Zookeeper if the race doesn't happen

consensus/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Repeatedly attempts to do a read modify write operation. Each attempt consists of a read of
+    /// the indicated node, a call to `f` with the data read (or None if the node did not exist),
+    /// and finally a write back to the node if it hasn't changed from when it was originally
+    /// written. The process aborts when a write succeeds or a call to `f` returns `Err`. In either
+    /// case, returns the last value produced by `f`.
+    fn read_modify_write<F, P, E>(&self, key: &str, f: F) -> Result<P, E>
+    where
+        F: FnMut(Option<P>) -> Result<P, E>,
+        P: Serialize + DeserializeOwned;
```
ms705: I guess these have to be here because the trait is intended to force any implementer to support these methods. However, unlike the other three above, they're not really part of a "public" API of an `Authority`, right? (In other words, I'm wondering if they would ever be called from outside the authority implementation.)

EDIT: ah, I see, the controller uses it to store other data in the authority.

consensus/src/zk.rs
```diff
@@ -0,0 +1,179 @@
...
+use Epoch;
+use CONTROLLER_KEY;
+
+struct EventWatcher;
+impl Watcher for EventWatcher {
+    fn handle(&self, e: WatchedEvent) {
+        if e.keeper_state != KeeperState::SyncConnected {
+            eprintln!("Lost connection to ZooKeeper! Aborting");
+            process::abort();
+        }
```
ms705: OOI, what's the purpose of `EventWatcher`? It doesn't do anything except check that the event's `KeeperState` is still `SyncConnected`.

EDIT: looks like it's used as an error handler for connecting to Zookeeper. Perhaps we should call it `ConnectSuccessWatcher` or something similar? I was misled into believing it was a more general event handler.

consensus/src/zk.rs
```diff
@@ -0,0 +1,179 @@
...
+        );
+        Self { zk }
+    }
+}
+
+impl Authority for ZookeeperAuthority {
+    fn become_leader(&self, payload_data: Vec<u8>) -> Epoch {
+        loop {
+            if let Some(epoch) = self.zk
+                .create(
```
ms705: Checking my understanding: we currently only support a single Soup deployment running against the same Zookeeper, i.e., there are no unique identifiers for different deployments. Correct?

I ask partly because it'd be nice to support multiple deployments, and partly because one could imagine that leftover Zookeeper state from a previous run is still around at this key as the lease hasn't expired yet, but we've already restarted Soup. I suspect the latter isn't a problem as we include a nonce in the payload (IIRC), so we won't be misled into believing that the dead leader is the current controller. Correct?

consensus/src/zk.rs
```diff
@@ -0,0 +1,179 @@
...
+                thread::park_timeout(Duration::from_secs(60));
+            }
+        }
+    }
+
+    fn await_new_epoch(&self, current_epoch: Epoch) -> Epoch {
+        loop {
+            match self.zk.exists_w(CONTROLLER_KEY, UnparkWatcher::new()) {
+                Ok(ref stat) if stat.czxid > current_epoch.0 => return Epoch(stat.czxid),
+                Ok(_) | Err(ZkError::NoNode) => thread::park_timeout(Duration::from_secs(60)),
```
ms705: This means we check for a new epoch every 60s, right?

src/controller/mod.rs
```diff
@@ -78,56 +85,18 @@ impl WorkerStatus {
...
-            sharding: None,
-            domain_config: DomainConfig {
-                concurrent_replays: 512,
-                replay_batch_timeout: time::Duration::from_millis(1),
-                replay_batch_size: 32,
-            },
-            persistence: Default::default(),
-            materializations: migrate::materialization::Materializations::new(&log),
+            config: ControllerConfig::default(),
             listen_addr: "127.0.0.1".parse().unwrap(),
```
ms705: Should this come from an argument? `src/controller/main.rs` has a parameter `address` that gets this information from the command line, but it doesn't seem to be passed into the `ControllerBuilder` anywhere.

EDIT: There's a `set_listen_addr` method below, but the `main.rs` doesn't call it -- should it do so?

src/controller/mod.rs
```diff
@@ -1482,124 +1354,545 @@ impl Drop for ControllerInner {
...
+            nworkers: 0,
+        }
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+struct ControllerState {
+    config: ControllerConfig,
+    epoch: Epoch,
+    recipe: (), // TODO: store all relevant state here.
```
ms705: This TODO doesn't affect the ability to run, because `ControllerInner` still stores the recipe in its `recipe` field, correct? AIUI, leaving this blank just means we don't persist the recipe across controller failures yet.

src/controller/mod.rs
```diff
@@ -1482,124 +1354,545 @@ impl Drop for ControllerInner {
...
+    ) -> JoinHandle<()> {
+        let descriptor = serde_json::to_vec(&descriptor).unwrap();
+        let builder = thread::Builder::new().name("srv-zk".to_owned());
+        builder
+            .spawn(move || {
+                loop {
+                    // become leader
+                    let current_epoch = authority.become_leader(descriptor.clone());
+                    let state = authority.read_modify_write(
+                        "/state",
```
ms705: Is it deliberate that this doesn't write the nonce from the descriptor as part of the state? I don't quite follow how we check it if it's not stored in Zk. Or am I misunderstanding its role?

Cargo.toml
```diff
@@ -79,6 +74,10 @@ path = "src/lib.rs"
 [[example]]
 name = "basic-recipe"
 
+[[bin]]
+name = "controller"
+path = "src/controller/main.rs"
```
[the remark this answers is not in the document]
fintelia: I agree that it probably makes sense to merge the two at some point, but currently I think that the `controller` binaries only run controllers and that `souplet` binaries are the only ones that run worker pools (with a small exception regarding a local worker poll for the currently elected leader)

consensus/src/lib.rs
fintelia: Yeah, the API of `Authority` was kind of tricky to figure out since I wanted it to live in a separate sub-crate without knowledge of Controller specific datatypes.

consensus/src/zk.rs
fintelia: This function is called with any out of band Zookeeper events that don't have designated event handlers specified for them. I'm currently only using it to detect if the connection to Zookeeper was lost and thus any associated leases expired (It is kind of unclear to to actually handle this case, so for now I just abort instead).

As a side note I think a failed connection would actually trigger an error earlier.

src/controller/mod.rs
fintelia: Exactly.

src/controller/mod.rs
fintelia: Good catch

jonhoo:
I just noticed: we should just use `thread::sleep` instead of `thread::park_timeout` when we're not explicitly calling `unpark` on the thread.

fintelia:
@jonhoo Agreed, but I don't see where that would apply

src/controller/mod.rs
fintelia: Wait, yes it does: https://github.com/mit-pdos/distributary/blob/39a6b0a28cb84f1087068db229256a39eda65229/src/controller/main.rs#L36

PR #53, opened 2017-12-08
Replace --unsharded with --shards in vote-stress

ekmartin:
Not sure if this is the only thing that needs fixing with vote-stress, however it compiles (and runs) now at least.


ms705:
Whoops, my bad -- I hadn't realized that `vote-stress` would be affected by the change in 3c001c147ff6e7b3b8c373cd5b46b00f5a5eda88.

Thanks for fixing :+1:

PR #55, opened 2018-01-24
Fix test failures

ekmartin:


jonhoo:
I think you also need to bump the version of `arccstr` to 1.0.4 for the current nightly

ekmartin:
Done!

ms705:
Looks good to me! :+1: 

@fintelia, can you double-check that all the new `unwrap`s added in the consensus code are in places where we actually want to crash on failure? I think so, but another pair of eyes would be good.

jonhoo:
Also, this still doesn't seem to pass?

ekmartin:
@jonhoo: it fails because of build errors in `benchmarks`. Didn't think it was worth fixing since it seemed like you were doing a pretty significant refactor in the `orchestrator` branch. Could add each subcrate manually with `--manifest-path` in the meantime though, at least then the tests would run.

jonhoo:
Ah, yes, fair. I haven't done anything to the `hotsoup` binary, but you're right, that seems fine. Once @fintelia signs off, I'd be happy for this to be merged, and then I'll merge it into `orchestrator` (which tbh should probably become `master` soon).

fintelia:
These changes look good to me. Lets merge and then just double check that all the tests do in fact work.

jonhoo:
@ekmartin can you rebase on top of master? For `Cargo.lock`, just remove the file, run `cargo update`, and commit the resulting one.

ekmartin:
@jonhoo: rebased 👍

jonhoo:
:tada:
@ms705 can you take a look at the `hotsoup` failures? Maybe we should just delete that binary from `master`?

PR #56, opened 2018-02-08
Remove window support from TCP channels

fintelia:
Our TcpSender/TcpReceiver supported running in a bounded mode, although this feature was almost never used. This PR removes that code

### DIFF
--- a/dataflow/src/domain/mod.rs
+++ b/dataflow/src/domain/mod.rs
@@ -173,8 +173,8 @@ impl DomainBuilder {
 
         let debug_tx = self.debug_addr
             .as_ref()
-            .map(|addr| TcpSender::connect(addr, None).unwrap());
-        let control_reply_tx = TcpSender::connect(&self.control_addr, None).unwrap();
+            .map(|addr| TcpSender::connect(addr).unwrap());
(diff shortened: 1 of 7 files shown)


dataflow/src/domain/mod.rs
```diff
@@ -1148,7 +1148,7 @@ impl Domain {
                                     use itertools::Itertools;
 
                                     let mut chunked_replay_tx =
-                                        TcpSender::connect(&domain_addr, Some(1)).unwrap();
+                                        TcpSender::connect(&domain_addr).unwrap();
```
jonhoo: So I guess this means we're assuming this being bounded doesn't matter.

jonhoo:
In essence this is saying that you should use `RpcChannel` if you want explicit acks.

dataflow/src/domain/mod.rs
jonhoo: Which is a *little* questionable, because your chunked replay is going to be sent all at once. Though maybe that's fine. You'll fill up the TCP buffers given that the reader reads round-robin.

jonhoo:
K, just going to merge this.

PR #58, opened 2018-02-16
[not for merging] SQlite Base Node Indices

ekmartin:
This implements persistent indices for Soup base nodes using SQlite, as a baseline for future experiments.

### Overview
At an API level it converts `State` into an enum that can be either `MemoryState` or `PersistentState` - with the latter now being used for base nodes. 

### Storage
Each instance of `PersistentState` corresponds to one table in SQlite. The table contains one column for each index and a single `row` column that stores a serialized representation of the `Vec<DataType>`. This might not be the right approach, so let me know if you have any better ideas!

#### Adding indices
`add_key` makes sure that there's a separate `index_#` column for all the parts of the given `columns` and finally adds a combined index using `CREATE INDEX`. This means that any rows that were inserted prior to `add_key` being called won't contain the new index columns though, so might be worth going through and backfilling those.

#### Look-ups
`lookup` builds up a `SELECT` query using the indices available in `columns`, on the form of `SELECT row FROM table_name WHERE index_0 = x AND index_1 = x AND ....`.

#### Removals
`remove` plucks out all the index values from the given row and performs a `DELETE FROM` with those.

#### Inserts
`insert` finds all the index values from the given row (from `self.indices`) and inserts them into separate `index_#` columns. The whole row gets put into a `row` column, after being serialized to JSON. This could probably be a `BLOB` column and use `bincode` instead as well.

### Future work
This mostly just implements the API - haven't run any benchmarks yet. This also performs an `INSERT` on every batch update, which with autocommit probably causes a disk write. @jonhoo suggested earlier that a possible solution might write on every update, but only flush to disk before the batch packet is inserted into the graph. Maybe this could be implemented by creating a transaction, and only committing it prior to insertion into the graph - not sure yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ run.sh
 *-log-*.json
 *.png
 *.log
+soup.db
 plotting/*.png
 
 callgrind.out.*
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,8 +331,10 @@ dependencies = [
  "petgraph 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "rahashmap 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rusqlite 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.27 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.27 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -886,6 +888,15 @@ version = "0.2.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "libsqlite3-sys"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "pkg-config 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "vcpkg 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "libssh2-sys"
 version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -909,6 +920,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "linked-hash-map"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "log"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -922,6 +938,14 @@ version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "cfg-if 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "lru-cache"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "linked-hash-map 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1498,6 +1522,17 @@ dependencies = [
  "hyper 0.10.13 (registry+https://github.com/rust-lang/crates.io-index)",
  "rusoto_core 0.31.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "xml-rs 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "rusqlite"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bitflags 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libsqlite3-sys 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lru-cache 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "time 0.1.39 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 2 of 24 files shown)


jonhoo:
Maybe I'm misremembering, but I think lookups should only ever use exactly one index. I don't think you should ever end up having to `AND` them together. You'd have to look at the `add_key`  for the old in-memory `State` to see whether it expects old rows to also be automatically be added to new indices. It probably does if something is a full materialization.

Some things to be aware of if you want this to also work for partial materalizations.
 - When a new index is added, it may be that existing rows do not need to be added to that index eagerly. In fact, I'm fairly sure that's not the case. But you should check the code.
 - Be aware that (for partial) you *need* to store empty results for keys too! That is, you need to be able to distinguish between `lookup(3)` returning `Missing` and returning no rows. The former requires a replay, the latter does not.

ekmartin:
@jonhoo: probably described it poorly - what I meant to say was that lookups `AND` together the columns of the given index, i.e. the `columns` argument given to `lookup`. That way it works for both single column and combined indices. I added a test with lookups against different indices as well [here](https://github.com/ekmartin/distributary/blob/95db92006ef59f490a6e3f3c2940c8cb37d3b2ee/core/src/local.rs#L1439-L1465).

I added a small fix that makes sure existing rows work with new indices as well.

Haven't looked into making it work with partial indices yet, but those are good points if I eventually need this to work with other nodes than bases.

ekmartin:
# Write performance
Here's an update regarding the raw write performance of the methods I've tried so far. All the benchmarks were run on `jerry`. For reference, here's how Soup performs with a write-only load normally:

4-shard-batch
4-shard-sjrn

## Replacing Soup's log with SQlite's write-ahead log
I had a theory that the performance gains from a possibly more optimized write-ahead log in SQlite could offset the reduced performance from having to persist data in SQlite. This would mean fsyncing every merged packet to SQlite, and only sending ACKs to clients after that has completed. 

While some initial results from mixed-load benchmarks indicated that this might be plausible, proper write-only benchmarks quickly showed that this was far from true. No matter the optimizations, I was never able to push the `vote` throughput to anything higher than ~45k ops/s on a single shard. In comparison, a pure write benchmark with Soup's log easily handles half a million ops/s.

Part of this is the indices' fault, as SQlite keeps each of its indices up to date on each write, even when using a WAL. This shows up clearly in flame graphs (zoom in on `worker1`):

* [Without indices](https://ekmartin.com/flame_graphs/wal.svg)
* [With indices](https://ekmartin.com/flame_graphs/wal_no_indices.svg)

While still slower than regular Soup, the performance is definitely better without any indices. Regardless, we'll need those indices for reading.

### Checkpointing

Another issue was the checkpoints SQlite was doing to move data from the WAL to the main database. By default this happens after the WAL reaches a threshold of a 1000 pages. This inflicts a latency hit on the commit that causes the threshold to be reached, which in `vote` sometimes amounted to up to **300ms** after the database started gaining a decent amount of rows. 

The SQlite docs explain the trade-off between read performance and how often checkpoints are taken. Since reads have to look up data in both the WAL and the main database, a larger WAL results in slower reads. From measuring insert and commit latencies in `vote` this seems to be the case for writes too, and this might be the reason the earlier mixed-load benchmarks performed much better: the write-load was never high enough to grow neither the WAL nor the main database to a significant size.

#### Checkpointing after ACKs

We know we have to checkpoint the WAL at some point, but we definitely don't have to do it before responding to clients. This meant disabling SQlite's automatic checkpoints, and instead manually checkpointing after a certain row limit was reached. 

This helped somewhat, and increased the throughput from a measly 25k ops/s to at least 42k ops/s. It doesn't really solve the problem though: checkpoints are still synchronous, and they still halt a domain's processing for a significant period. The time it takes to create a checkpoint grows gradually with the database size: initially a couple of milliseconds, but up to hundreds towards the end of `vote`. The latency does depend on how often a checkpoint is taken, however even with checkpoints every 10k rows, the later checkpoints ended up taking at least a 100ms. As a comparison, the time it takes to persist a merged packet of ~400 rows to SQlite starts out at less than a millisecond and grows to around 10ms.

#### Checkpointing in a separate thread

The SQlite docs mention the possibility of doing manual checkpoints in a separate thread. The [`sqlite3_wal_checkpoint`](https://www.sqlite.org/c3ref/wal_checkpoint_v2.html) method comes in a few flavors, but whereas the `PASSIVE` mode might seem promising, as it doesn't lock the database - this only helps if the write throughput is low enough to allow "breaks" between writes where these checkpoints might happen. If that's not the case the checkpoint calls will just be no-ops, and the WAL will continue to grow. 

The other modes would instead block write processing, which would basically be the same as having automatic checkpoints on (since we only write from one thread per database).

## Without fsyncing to SQlite

So, back to the original idea: writing to Soup's log first, and then only persisting data to SQlite after sending ACKs to clients. This means that we can relax the durability guarantees of SQlite somewhat. The benchmarks here use the following:

[`synchronous = OFF`](https://www.sqlite.org/pragma.html#pragma_synchronous) - never wait for SQlite to fsync data before returning
[`journal_mode = OFF`](https://www.sqlite.org/pragma.html#pragma_journal_mode) - don't keep a journal at all, effectively disabling atomic
commit/rollback

This is not very optimal in SQlite's case, as a poorly timed failure could cause the database to go corrupt. `journal_mode = WAL` would probably be better in that regard, and with `synchronous = NORMAL` it would only fsync prior to checkpoints happening. However, the more relaxed guarantees are probably closer to what we ultimately want to achieve with a persistent index structure when we're syncing the log before ACKing every write (as long as we can make it work with recovery). It's also a more useful comparison point: if this is still too slow, then the stricter guarantees are not going to work either.

### Results

Unfortunately, while the results are better than in the previous section, it is
still really slow. With a single shard the throughput never pushes past 100k
ops/s, with quickly growing latencies.

1-shard-batch
1-shard-sjrn

With 4 shards the throughput climbs to about 300k ops/s, but the latencies are still way too high.

4-shard-batch
4-shard-sjrn

### Batching

In the implementation above, each merged packet results in a single SQlite transaction.
Since it seemed like committing these transactions was taking a significant amount of time, I thought that perhaps batching up a larger amount of rows into each transaction would be an improvement. This proved to be wrong, as it seems like the commit time scales somewhat linearly past a specific amount of rows, which in hindsight makes sense. Inserting 10k rows takes ~100ms (+ 5ms to commit), while 100k rows takes about 1000ms (+ 50ms to commit).

This means that the only thing it really helps with is the 50th percentile remote latency, for those few packets that are lucky enough not to go past the batch limit. This doesn't really help throughput at all though, as the batches still happen on the regular domain path (in fact, the throughput goes down slightly). 

4-shard-batch-batch
4-shard-batch-sjrn

Code: https://github.com/mit-pdos/distributary/compare/master...ekmartin:base_indices_batched

### SQlite Conclusion

The fact that SQlite works quite well out of the box without a lot of tuning is definitely one of its strengths. Regardless, there's a few knobs that can turned both at runtime and compile time.

I experimented with a few of the recommended compile options [here](https://www.sqlite.org/compile.html), such as completely removing the mutex code and disabling unnecessary extra features. Although I didn't get much of an impact out of this, I'm sure there are options that could've improved the situation a little. Regardless, I don't think that's where the main problem lies, and at this point I suspect that no matter what, keeping the necessary indices up to date on the main path with SQlite is going to be too slow.

Does this mean that directly maintaining something like B-tree indices on the main materialization path would perform too poorly as well? Possibly. SQlite definitely has a lot of overhead we don't really need, but in the end, the bulk of the time is spent maintaining its B-trees (as [this flame graph](https://ekmartin.com/flame_graphs/no_wal.svg) shows).

## RocksDB

As an experiment, I thought I'd see how something like RocksDB fared in terms of raw write performance. RocksDB normally works by persisting writes to a WAL, while holding data in a memtable structure, which later on gets flushed to disk in the form of SSTables. We already have a WAL though, so instead of writing to a WAL again, we'll just write to memtables and have RocksDB's background threads eventually flush the data to disk. Not a very fair comparison to SQlite, which would have to do all the writing on the main materialization path, but still useful as a benchmarking point.

Whereas SQlite's defaults are quite okay, RocksDB potentially requires a lot of tuning, which I haven't gotten around to yet. The initial results are decent though, maintaining a throughput well above 500k ops/s:

4-shard-rocksdb-batch
4-shard-rocksdb-sjrn

The question then however, is the read performance. I'll post an update once I have a decently tuned benchmark up and running, but I fear that it would be worse than SQlite's (which uses B-trees), simply from the fact that it uses LSM-trees as its data structure.

Code: https://github.com/mit-pdos/distributary/compare/master...ekmartin:rocksdb_base_indices

## Tailing the Soup log

This was what @ms705 proposed when we initially discussed persistent base indices.

If we're not going to write to disk directly on the main materialization path, it might be better to do the background processing ourselves instead of relying on something like RocksDB to do so. A set of log files could be assigned to a persistence worker, which would read and persist all new entries to SQlite in batches.

As in the SQlite batching implementation, reads would first refer to SQlite, and then a regular `MemoryState` cache - in case any relevant rows were inserted after the last flush. Keeping all this synchronized across different threads will definitely add some complexity, but it might be the best solution regardless - especially if my assumptions regarding RocksDB read performance turn out to be correct. 

I'll try implementing it in the next few days, and post an update after doing so.

ekmartin:
Closing in favor of #72.

PR #59, opened 2018-02-27
Use multi_put for benchmark prepopulation

ekmartin:
Doesn't matter too much, but saves a little waiting time.

### DIFF
--- a/benchmarks/vote/clients/localsoup/mod.rs
+++ b/benchmarks/vote/clients/localsoup/mod.rs
@@ -55,9 +55,10 @@ impl VoteClient for Client {
             println!("Prepopulating with {} articles", params.articles);
         }
         let mut a = g.graph.get_mutator("Article").unwrap();
-        a.batch_put(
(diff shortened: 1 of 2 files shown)


jonhoo:
I think we actually specifically want `batch_put` here, because otherwise we generate a *single* potentially *very* large update (e.g., if you used `-a 10000000`). Maybe the answer is to do `multi_put` inside of the `batch_put`... I'm surprised to hear that `multi_put` is significantly faster than `batch_put`, as they should both just fill up TCP buffers...

ekmartin:
Closing this now that `batch_put` doesn't hang.

jonhoo:
I think ultimately we'll want `batch_put` + `multi_put`. if a `batch_put` runs for too long without checking for acks, it could totally generate a deadlock. The domain of the base will fill the TCP buffer trying to send more acks, but that buffer will never drain because the sender will block on the domain not reading from the base input. The alternative is to make `batch_put` occasionally read acks.

PR #61, opened 2018-03-13
Don't send empty packets to wrong shards

ekmartin:
Previously base domain shards that didn't match `shard_by` for a given key would still get a packet - just without any records in it.

### DIFF
--- a/src/controller/domain_handle.rs
+++ b/src/controller/domain_handle.rs
@@ -59,7 +59,8 @@ impl<'a> BatchSendHandle<'a> {
             }
             let key_col = key[0];
 
-            let mut shard_writes = vec![Vec::new(); self.dih.txs.len()];
+            let mut shard_writes: HashMap<usize, Vec<Record>> =
+                HashMap::with_capacity(self.dih.txs.len());
             let mut data = p.take_data();
             for r in data.drain(..) {
                 let shard = {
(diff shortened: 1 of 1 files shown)


src/controller/domain_handle.rs
```diff
@@ -59,7 +59,8 @@ impl<'a> BatchSendHandle<'a> {
             }
             let key_col = key[0];
 
-            let mut shard_writes = vec![Vec::new(); self.dih.txs.len()];
+            let mut shard_writes: HashMap<usize, Vec<Record>> =
+                HashMap::with_capacity(self.dih.txs.len());
```
jonhoo: I think I'd prefer for this to say a `Vec<Vec<_>>`. Any particular reason you want it to be a `HashMap`? I guess if there are *a lot* of shards?

jonhoo:
As long as this doesn't make any tests/benchmarks break (that weren't before), I'd be happy to merge this.

src/controller/domain_handle.rs
```diff
@@ -69,10 +70,10 @@ impl<'a> BatchSendHandle<'a> {
                     };
                     dataflow::shard_by(key, self.dih.txs.len())
                 };
-                shard_writes[shard].push(r);
+                shard_writes.entry(shard).or_insert_with(Vec::new).push(r);
             }
 
-            for (s, rs) in shard_writes.drain(..).enumerate() {
```
fintelia: Instead of using a HashMap, you could just replace this line with:

```
for (s, rs) in shard_writes.drain(..).filter(|rs| !rs.is_empty()).enumerate() {	
```

ekmartin:
@jonhoo: the `HashMap` was just a micro-optimization to avoid having to always create empty vectors for all shards. I changed it to just use an `if` instead - seems to pass the same tests (and `vote` still works).

src/controller/domain_handle.rs
ekmartin: I think that would make the indices wrong? Since you're filtering out before you call `.enumerate()`. And filtering afterwards probably wouldn't work too well in terms of the borrow checker right (since you'd need to move out `rs`)?

jonhoo:
`Vec::new` actually does *no* allocation, so it's really cheap to create :)

ekmartin:
Ah I see, that's good to know!

PR #64, opened 2018-03-15
Remove generics from State

ekmartin:
We discussed this a while back, and I ended up implementing it for #58. If this still sounds like a good idea it'd be nice to get it into master to avoid having to fix conflicts every time something in `core` changes.

### DIFF
--- a/core/src/local/state.rs
+++ b/core/src/local/state.rs
@@ -5,20 +5,11 @@ use std::rc::Rc;
 use local::keyed_state::KeyedState;
 use local::single_state::SingleState;
 
+#[derive(Default)]
 pub struct State {
     state: Vec<SingleState>,
     by_tag: HashMap<Tag, usize>,
     rows: usize,
-}
-
-impl Default for State {
-    fn default() -> Self {
-        State {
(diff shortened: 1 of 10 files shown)


core/src/local/state.rs
```diff
@@ -1,18 +1,17 @@
...
-pub struct State<T: Hash + Eq + Clone + 'static> {
-    state: Vec<SingleState<T>>,
+pub struct State {
+    state: Vec<SingleState>,
     by_tag: HashMap<Tag, usize>,
     rows: usize,
 }
 
-impl<T: Hash + Eq + Clone + 'static> Default for State<T> {
+impl Default for State {
```
fintelia: We could actually `#[derive(Default)]` for State now that the bounds don't get in the way

ekmartin: Nice! Fixed in https://github.com/mit-pdos/distributary/pull/64/commits/59fd0c23f8e52b87e87c30a78e5184d0b8155d89.

jonhoo:
Yeah, I like this!

PR #67, opened 2018-03-18
Fix core tests

ekmartin:
Just two tiny `as u64` casts that were missing for the `core` tests to pass.


ekmartin:
Removed `serde_json` from `core` as well, since it doesn't seem like it's being used there anymore.

jonhoo:
LGTM.

PR #68, opened 2018-03-21
Upgrade nom

ekmartin:
So that I can use it in distributary-mysql :)


ekmartin:
Hold on, I'll try to see if there's some code that can be simplified now as well.

ms705:
I think you found the code that can be removed. Ready to merge?

ekmartin:
Yep, should be good to go!

PR #69, opened 2018-03-25
Skip flaky shuffle tests

ekmartin:
Also upgrades the nightly version on Travis - the old one has a compiler bug that has since been fixed (https://travis-ci.org/mit-pdos/distributary/builds/358092643?utm_source=github_status&utm_medium=notification).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: trusty
 language: rust
 rust:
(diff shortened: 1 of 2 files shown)


ms705:
Seems good! We shouldn't forget about these tests though, as getting them to pass eventually is important.

PR #70, opened 2018-03-25
`&slog::Logger` -> `slog::Logger`

ekmartin:
This caused crashes on Travis in `distributary-mysql` previously: https://travis-ci.org/ekmartin/distributary-mysql/builds/358136589#L1341


ekmartin:
Seems like this is caused by https://github.com/slog-rs/slog/issues/164 :)

PR #71, opened 2018-03-26
Upgrade slog to 2.2.2

ekmartin:
This should solve the issue we were having with `distributary-mysql`. Also upgraded slog-term to the latest version.


ekmartin:
The fix didn't include `&mut Logger` so I removed the only location where we used that in https://github.com/mit-pdos/distributary/pull/71/commits/8a83bb79fcf798ee3e8791c98fc5922d370b4d81. Commented on the slog issue though.

ekmartin:
For some reason Travis didn't want to build the second commit, but it passed on my fork at least: https://travis-ci.org/ekmartin/distributary/builds/358637939

PR #72, opened 2018-04-07
RocksDB Persistence

ekmartin:
## Overview
This implements `PersistentState` from #58 with RocksDB instead of SQlite.

## Storage
Similar to how the SQlite implementation worked, each `PersistenceState` opens its own RocksDB database. Since RocksDB is a key-value store the implementation differs quite a bit though. The scheme I've implemented so far is definitely far from perfect, but here's how it looks so far:

### Keys
Keys are tuples on the form of `(KeyIndex, Vec<DataType>, KeySeq)`:
* `KeyIndex` signifies the index in `self.indices` and lets us include only values for the relevant columns in the key. Consider the table `(a: int, b: int, c: int)` and the indices `(a, b)` and `(a, c)` - including the `KeyIndex` here lets us separate a key for the first index from a key from the second.
* `Vec<DataType>` is the actual key value
* `KeySeq` is a monotonically increasing sequence number that is currently maintained in memory for each index, and lets us write multiple values for each key without having to retrieve the existing rows for that key first.

#### Primary Index
The first index added through `add_key` is considered a primary index, and is responsible for storing the actual row values inserted into state. With the example table from the last section, an insert of the row `(1, 2, 3)` would result in a `(0, (1, 2), 0)` key pointing to the row. 

Note that it makes no assumption on whether the primary index should be unique or not, which is an optimization that might be interesting to look into later: if we know that the first index is a primary key, we wouldn't need to support storing multiple values for each key.

#### Secondary Indices
The subsequent indices always point to the primary key. With the same example table, an insert of `(1, 2, 3)` would cause our second index to insert the key `(1, (1, 3), 0)` pointing to the primary key `(1, 2)`.

### Lookups
All lookups rely on a RocksDB feature called prefix seeks to be able to maintain multiple values per Soup key. We construct a key prefix on the form of `(KeyIndex, Vec<DataType>)` and iterate through the values for that prefix.

For primary indices that's all there is. Secondary indices require another lookup for each primary key. We could avoid this by always storing the complete row for all indices, but that would require writing and storing more bytes.

### Serialization
Both keys and values are currently serialized with `bincode`.

### Constructing prefixes
To be able to perform the prefix seeks described in the lookup section we need to be able to construct prefixes from the serialized bytes. This should ideally be done without any deserialization or serialization at all, however it currently requires deserializing a part of the key - which is definitely a potential source of improvement: https://github.com/ekmartin/distributary/blob/f3ed68d988157e00ccb1306e90d48dfebb8bdb2c/core/src/local/state.rs#L199

### Future work
MyRocks, Facebook's MySQL implementation on top of RocksDB, would definitely be a good source of inspiration if we choose to go down this path. The way they build up what they call memcomparable keys would probably be a good idea to try to replicate: https://github.com/facebook/mysql-5.6/wiki/MyRocks-record-format

## Persistence
Writes are written to Soup's log in the same way as normal, and RocksDB's write-ahead log has been disabled. Writes are first written to RocksDB's memtables (in-memory), which are later flushed to disk by RocksDB background threads (the `--persistence-threads` argument to `vote`). Note that these threads are shared across all RocksDB instances for the same process, so even if we have 4 shards and enable 4 background threads for each of them, that will still result in a total of 4 RocksDB background threads.

## Results
Benchmarks were run on `jerry`. 

### Write-only vote
* 4 shards
* 16 workers
* 10 client threads
* 4 RocksDB background threads

plot-batch
plot-sjrn

### 95% reads vote with no warmup
* 4 shards
* 4 workers
* 16 read threads
* 4 RocksDB background threads

This is a less interesting comparison than the write-only benchmark, simply because of Soup's caching nature. Whereas the first reads will trigger replays, and thus hit RocksDB, subsequent reads will mostly use Soup's `evmaps`. 

plot-batch
image

### Future work
As @ms705 pointed out on Slack:
> for reads, you'll want to force *every* request to be a replay

> so I would either (a) modify the benchmark to read each key only once
> or (b) force immediate eviction of the keys read (a bit suboptimal, as it muddles read cost with the extra cost of evictions flowing through the graph)

This would let us benchmark the read performance of *only* RocksDB compared to our regular state implementation. After we have that it'd be useful to run a real-world benchmark with persistent states and eviction.

## Tuning RocksDB
RocksDB has a lot of potential for tweaking. So far I've mostly been experimenting with the bigger architectural options:

### Memtable format
The default RocksDB memtable is a skiplist. Since we're using prefixes for all reads, there's a lot of improvement to be found in changing this to one of the [hash based implementations](https://github.com/facebook/rocksdb/wiki/MemTable), which places values in buckets based on their prefix.  Changing from a skip list to a hash skip list resulted in a `~150k ops/s` increase in write-only, and a second change to using a hash linked list further increased this with `~50k ops/s`. 

The comparison between the two hash implementations mostly comes down to the number of values for each keys. For really large secondary indices this means that a hash skip list would probably serve better, so we might want to revisit this in the future.

### SS-table index type
As described [here](https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide#prefix-databases), a similar trade-off exists for the on-disk SS-tables. Enabling a hash index here didn't have as much of an impact though, but it might help for larger amounts of data - will have to re-visit.

### DIFF
--- a/benchmarks/replay/main.rs
+++ b/benchmarks/replay/main.rs
@@ -5,6 +5,7 @@ extern crate hdrhistogram;
 extern crate itertools;
 extern crate rand;
 
+use std::fs;
 use std::time::{Duration, Instant};
 use std::path::PathBuf;
 use std::process::Command;
@@ -65,7 +66,16 @@ fn populate(g: &mut ControllerHandle<ZookeeperAuthority>, rows: i64, verbose: bo
         });
 }
 
-fn perform_reads(g: &mut ControllerHandle<ZookeeperAuthority>, reads: i64, rows: i64) {
+fn perform_reads(
+    g: &mut ControllerHandle<ZookeeperAuthority>,
+    reads: i64,
+    rows: i64,
+    verbose: bool,
+) {
+    if verbose {
+        eprintln!("Done populating state, now reading articles...");
+    }
+
     let mut hist = Histogram::<u64>::new_with_bounds(1, 100_000, 4).unwrap();
     let mut getter = g.get_getter("ReadRow").unwrap();
 
@@ -114,23 +124,30 @@ fn main() {
                 .help("Number of rows to read while benchmarking"),
         )
         .arg(
-            Arg::with_name("use-existing-data")
-                .long("use-existing-data")
-                .requires("retain-logs-on-exit")
-                .takes_value(false)
-                .help("Skips pre-population and instead uses already persisted data."),
-        )
-        .arg(
             Arg::with_name("log-dir")
                 .long("log-dir")
                 .takes_value(true)
                 .help("Absolute path to the directory where the log files will be written."),
         )
         .arg(
+            Arg::with_name("durability")
+                .long("durability")
+                .takes_value(false)
+                .help("Enable durability for Base nodes"),
+        )
+        .arg(
             Arg::with_name("retain-logs-on-exit")
                 .long("retain-logs-on-exit")
                 .takes_value(false)
+                .requires("durability")
                 .help("Do not delete the base node logs on exit."),
+        )
+        .arg(
+            Arg::with_name("use-existing-data")
+                .long("use-existing-data")
+                .requires("retain-logs-on-exit")
+                .takes_value(false)
+                .help("Skips pre-population and instead uses already persisted data."),
         )
         .arg(
             Arg::with_name("write-batch-size")
@@ -168,6 +185,7 @@ fn main() {
     assert!(reads < rows);
 
     let verbose = args.is_present("verbose");
+    let durable = args.is_present("durability");
     let flush_ns = value_t_or_exit!(args, "flush-timeout", u32);
 
     let mut persistence = PersistenceParameters::default();
@@ -175,6 +193,12 @@ fn main() {
     persistence.persistence_threads = value_t_or_exit!(args, "persistence-threads", i32);
     persistence.queue_capacity = value_t_or_exit!(args, "write-batch-size", usize);
     persistence.log_prefix = "replay".to_string();
+    persistence.mode = if durable {
+        DurabilityMode::Permanent
+    } else {
+        DurabilityMode::MemoryOnly
+    };
+
     persistence.log_dir = args.value_of("log-dir")
         .and_then(|p| Some(PathBuf::from(p)));
 
@@ -182,33 +206,30 @@ fn main() {
         args.value_of("zookeeper-address").unwrap(),
     ));
 
-    // Populate in a closure, so we'll have to recover before performing reads.
-    if args.is_present("durability") && !args.is_present("use-existing-data") {
+    if !args.is_present("use-existing-data") {
         let mut g = build_graph(authority.clone(), persistence.clone(), verbose);
         g.install_recipe(RECIPE.to_owned()).unwrap();
 
         // Prepopulate with n rows:
         populate(&mut g, rows, verbose);
+
+        // In memory-only mode we don't want to recover, just read right away:
+        if !durable {
+            perform_reads(&mut g, reads, rows, verbose);
+            return;
+        }
     }
 
-    persistence.mode = if args.is_present("durability") {
-        if args.is_present("retain-logs-on-exit") {
-            DurabilityMode::Permanent
-        } else {
-            DurabilityMode::DeleteOnExit
-        }
-    } else {
-        DurabilityMode::MemoryOnly
-    };
-
+    // Recover the previous graph and perform reads:
     let mut g = build_graph(authority, persistence, verbose);
     // Flush disk cache:
     Command::new("sync")
         .spawn()
         .expect("Failed clearing disk buffers");
-    if verbose {
-        eprintln!("Done populating state, now reading articles...");
-    }
+    perform_reads(&mut g, reads, rows, verbose);
 
-    perform_reads(&mut g, reads, rows);
+    // Remove any log/database files:
+    if !args.is_present("retain-logs-on-exit") {
+        fs::remove_dir_all("replay-TableRow-0.db").unwrap();
+    }
 }


ekmartin:
## With RocksDB's write-ahead log
Instead of relying on Soup's log for durability, it'd be nice if we could enable RocksDB's WAL and let it handle persistence completely. This would make recovery significantly easier, and faster, as RocksDB makes sure the WAL doesn't grow too large.

### Implementation
Each merged packet results in a RocksDB `WriteBatch` written synchronously (waits for the `fdatasync` call to return).

### Results
Both of the benchmarks were run with 4 RocksDB background workers and 10 client threads. 

#### With 1 shard (4 workers)
plot-batch
plot-sjrn

#### With 4 shards (16 workers)
plot-batch
plot-sjrn

#### Notes
It doesn't seem like the size of the `WriteBatch` matters too much after a certain threshold, which should indicate that the `fdatasync` cost isn't too significant. A packet of 2000 records takes about ~3 ms to write (1 shard), and this is pretty stable throughout the benchmark, with a write speed of about 1 - 2μs per record.

There are a few spikes throughout though, which might have a significant impact on the sojourn time. In those cases a packet of 2000 records might take up to almost 30ms to write (~14μs per record). In very rare cases it seems to be even worse: one 2600 record packet took 210ms to write! I'm not completely certain what causes the spikes yet, but it's probably due to the background processes flushing or compacting data at the same time as the WAL being written to disk.

An interesting comparison would be with a benchmark where the write-ahead log is written to a separate disk, which would make sure that the background threads never compete with disk capacity with the main thread writing to the WAL. As far as I've understood this is a fairly regular setup in production databases as well (https://www.postgresql.org/docs/9.2/static/wal-internals.html).

ekmartin:
## Writing the WAL to a separate disk
RocksDB writes to disk in at least three separate scenarios:
* Flushing the write-ahead log
* Flushing memtables to disk (high priority background thread)
* Compacting SS-tables on disk (low priority background thread)

In the previous comment I mentioned that the spikes in write latency might be a result of the main thread competing with disk resources with RocksDB's background threads. After running a couple benchmarks on `jerry` with the WAL being written to a separate disk, this seems to be true.

### Benchmarks
Same setup as in the previous comment, however the main database files now reside on a separate SSD (`/flash` on jerry). The log files are written to the same disk as in the last benchmarks (`/scratch` on jerry), both for the Soup log and the RocksDB log. If these were flipped the results would be much worse, as `/scratch` seems to be a faster disk than the one mounted to `/flash`. 

#### With 1 shard (4 workers)
plot-batch
plot-sjrn

#### With 4 shards (16 workers)
plot-batch
plot-sjrn

Not too bad! Interestingly, the RocksDB implementation actually beats the Soup log at lower throughput targets (here with 1 shard):

low-batch
low-sjrn

ekmartin:
@ms705: No, I think it should be ready for review! 

The only thing I can think of would be lookups of identical values with different types (e.g. insert `DataType::Int(1)` and retrieve by `DataType::BigInt(1)`) which is something `MemorySoup` supports. The best way to solve this might be to normalize keys somewhere during lookups, i.e. turn an `BigInt` into an `Int` if it actually fits in an `i32`.

ms705:
Hmm, good point. I think it's okay to merge this without -- though we'll have to be careful to use the right types for keys in our benchmarks. I don't see an easy way of implementing the same hack that we do in in-memory Soup (where `BigInt` and `Int` always compare equal if the value is equal).

Given the limited time until OSDI, let's prioritize the merge so that we can run experiments with the durability enabled. Is there extra setup required when using the new persistent bases? (e.g., does RocksDB have to run listening on a specific port?)

ekmartin:
Yep, that sounds good. And no, no ports needed - RocksDB is purely embedded. It does build RocksDB from source as a part of the build process though, and needs lz4 installed as a prerequisite (`liblz4-dev` on Debian and `lz4` on Arch).

There's also a couple of new `PersistenceParameters` options to be aware of, mainly `log_dir` and `persistence_threads`. The latter defines the amount of background threads used by RocksDB (for flushing and compacting) - I've been using 4 when running `localsoup` but depending on the write load a lesser amount might be fine as well.

core/src/lib.rs
```diff
@@ -21,10 +23,107 @@ pub mod data;
...
+}
+
+impl PersistenceParameters {
+    /// Parameters to control the persistence mode, and parameters related to persistence.
+    ///
+    /// Three modes are available:
+    ///
+    ///  1. `DurabilityMode::Permanent`: all writes to base nodes should be written to disk.
+    ///  2. `DurabilityMode::DeleteOnExit`: all writes are written to disk, but the log is
+    ///     deleted once the `Blender` is dropped. Useful for tests.
```
jonhoo: Should this say "writes *to base nodes*"?

dataflow/src/persistence/mod.rs
```diff
@@ -189,33 +116,36 @@ impl GroupCommitQueueSet {
...
-                    serde_json::to_writer(&mut file, &data_to_flush).unwrap();
-                    // Separate log flushes with a newline so that the
-                    // file can be easily parsed later on:
-                    writeln!(&mut file, "").unwrap();
-                }
+        // With persistent bases we're already maintaining a
+        // write-ahead log, so no need to log twice.
+        if !self.params.persist_base_nodes
+            && (self.params.mode == DurabilityMode::DeleteOnExit
+                || self.params.mode == DurabilityMode::Permanent)
```
jonhoo: I think we should just get rid of the old append-only log. It complicates the code, and we've already done the benchmarking. That should also let us stick with just `DurabilityMode` if I read the code right (I know *technically* it lets us stick with `persist_base_nodes`, but I'd prefer the `enum` and especially also to keep support for deleting the base storage on exit).

dataflow/src/domain/mod.rs
```diff
@@ -1446,15 +1450,16 @@ impl Domain {
         }
 
         let n = self.nodes[&source].borrow();
+        let data: &Vec<DataType> = &*row;
```
jonhoo: This should not be here (see below).

dataflow/src/domain/mod.rs
```diff
@@ -1446,15 +1450,16 @@ impl Domain {
...
         if n.is_internal() {
             if let Some(b) = n.get_base() {
-                let mut row = ((*row).clone(), true).into();
+                let mut row = (data.clone(), true).into();
```
jonhoo: `row.into_owned()`

dataflow/src/domain/mod.rs
jonhoo: This seems like an unnecessary change?

dataflow/src/node/process.rs
```diff
@@ -175,27 +175,23 @@ impl Node {
...
+                        Packet::Message {
                             ref mut senders, ..
-                        },
-                    ) => senders.drain(..).for_each(|src| ex.send_back(src, Ok(0))),
-                    (
-                        Some(..),
-                        Some(ex),
-                        &mut box Packet::Transaction {
+                        } => senders.drain(..).for_each(|src| ex.send_back(src, Ok(0))),
+                        Packet::Transaction {
```
jonhoo: This change also seems unnecessary unless I'm missing something?

dataflow/src/ops/base.rs
```diff
@@ -179,17 +179,20 @@ impl Ingredient for Base {
...
-                        1 => Some(Cow::Borrowed(&*rows[0])),
                         n => {
                             // primary key, so better be unique!
                             assert_eq!(n, 1, "key {:?} not unique (n = {})!", current_key, n);
                             unreachable!();
                         }
                     }
                 }
+                LookupResult::Some(Cow::Owned(mut rows)) => {
+                    Some(Cow::from(rows.pop().unwrap().unpack()))
```
jonhoo: Nice catch :tada:

dataflow/src/ops/filter.rs
```diff
@@ -170,42 +171,48 @@ impl Ingredient for Filter {
...
+                        // everything matches no condition
+                        true
+                    }
+                })
+            };
+
             match state.lookup(columns, key) {
-                LookupResult::Some(rs) => {
+                LookupResult::Some(Cow::Borrowed(rs)) => {
+                    let r = Box::new(rs.iter().filter(filter).map(|r| Cow::from(&r[..]))) as Box<_>;
```
jonhoo: Shouldn't this be `.into_iter` so that we can map with `Cow::Owned` and not have to eventually clone somewhere down the line?

core/src/local/mod.rs
```diff
@@ -16,10 +55,17 @@ impl Tag {
...
 pub struct Row(pub(crate) Rc<Vec<DataType>>);
 
 unsafe impl Send for Row {}
 
+impl Row {
+    pub fn unpack(self) -> Vec<DataType> {
+        Rc::try_unwrap(self.0).unwrap()
+    }
+}
+
```
jonhoo: I'm a little sad about the indirection through `Rc` for `PersistentState`, but I guess it's okay..

dataflow/src/ops/grouped/mod.rs
```diff
@@ -222,6 +222,7 @@ where
                         }
                     };
 
+                    let old = rs.get(0);
```
jonhoo: I feel like we can probably take advantage of `old` being `Cow::Owned` here too, but probably not worth spending time on. Maybe add a `TODO`?

dataflow/src/ops/join.rs
```diff
@@ -431,44 +431,44 @@ impl Ingredient for Join {
                     while other_rows.peek().is_some() {
                         if let Some(false) = make_null {
                             // we need to generate a -NULL for all these lefts
-                            ret.push((self.generate_null(other), false).into());
+                            ret.push((self.generate_null(&other), false).into());
```
jonhoo: Wouldn't it be better to have the various `generate_*` things take `Cow`s now so that they can just move out the `DataType`s rather than cloning them?

dataflow/src/ops/latest.rs
```diff
@@ -94,8 +94,8 @@ impl Ingredient for Latest {
             });
 
             // buffer emitted records
-            for (r, current) in currents {
-                if let Some(current) = current {
+            for (r, current_row) in currents {
+                if let Some(current) = current_row.get(0) {
                     out.push(Record::Negative((**current).clone()));
```
jonhoo: Can we get rid of this clone (or one further down) if `current_row` is a `Cow::Owned`?

dataflow/src/processing.rs
```diff
@@ -217,12 +218,16 @@ where
...
         domain: &prelude::DomainNodes,
         states: &'a prelude::StateMap,
-    ) -> Option<Option<Box<Iterator<Item = &'a [prelude::DataType]> + 'a>>> {
+    ) -> Option<Option<Box<Iterator<Item = Cow<'a, [prelude::DataType]>> + 'a>>> {
         states
             .get(&parent)
             .and_then(move |state| match state.lookup(columns, key) {
-                prelude::LookupResult::Some(rs) => {
-                    Some(Some(Box::new(rs.iter().map(|r| &r[..])) as Box<_>))
+                prelude::LookupResult::Some(Cow::Owned(rs)) => Some(Some(Box::new(
```
jonhoo: This file should probably `use prelude::*;`

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+    epoch: IndexEpoch,
+}
+
+impl PersistentIndex {
+    fn new(columns: Vec<usize>) -> Self {
+        Self { columns, seq: 0 }
+    }
+}
+
+/// PersistentState stores data in SQlite.
```
jonhoo: This comment is a lie.

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+pub struct PersistentState {
+    name: String,
+    db_opts: rocksdb::Options,
+    // We don't really want DB to be an option, but doing so lets us drop it manually in
+    // PersistenState's Drop by setting `self.db = None` - after which we can then discard the
+    // persisted files if we want to.
+    db: Option<DB>,
```
jonhoo: Can you used `rocksdb::DB` instead of `DB` here just to disambiguate?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+
+impl SizeOf for PersistentState {
+    fn size_of(&self) -> u64 {
+        use std::mem::size_of;
+
+        size_of::<Self>() as u64
+    }
+
+    fn deep_size_of(&self) -> u64 {
+        self.size_of()
```
jonhoo: This seems false? Though I guess we never want to explicitly evict from bases anyway? @ms705 ?

core/src/local/mod.rs
```diff
@@ -1,12 +1,51 @@
...
+    /// Add an index keyed by the given columns and replayed to by the given partial tags.
+    fn add_key(&mut self, columns: &[usize], partial: Option<Vec<Tag>>);
+
+    /// Returns whether this state is currently keyed on anything. If not, then it cannot store any
+    /// infromation and is thus "not useful".
+    fn is_useful(&self) -> bool;
+
+    fn is_partial(&self) -> bool;
+
+    fn process_records(&mut self, records: &mut Records, partial_tag: Option<Tag>);
```
jonhoo: I think it'd be useful to include a doc here about why we give `&mut Records`. The reason is specifically that if the state is partial and a given record misses in all indices in a state, then it should be *removed* from `records`. This is quite important, and I have completely forgotten!

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+                i == 0
+            })
+    }
+
+    // Puts by primary key first, then retrieves the existing value for each index and appends the
+    // newly created primary key value.
+    // TODO(ekmartin): This will put exactly the values that are given, and can only be retrieved
+    // with exactly those values. I think the regular state implementation supports inserting
+    // something like an Int and retrieving with a BigInt.
+    fn insert(&mut self, batch: &mut WriteBatch, r: Vec<DataType>) {
```
jonhoo: Why can't this take a `&[DataType]` so that we don't have to clone in `process_records`?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+                Record::Negative(ref r) => {
+                    self.remove(&mut batch, r);
+                }
+                Record::BaseOperation(..) => unreachable!(),
+            }
+        }
+
+        // Sync the writes to RocksDB's WAL:
+        let mut opts = rocksdb::WriteOptions::default();
+        opts.set_sync(true);
```
jonhoo: This seems really weird to me. Surely we should have already written to the WAL before we get to this point? Unless I'm misremembering GroupCommitQueue, acks are sent once the flush method there is called (which does nothing for RocksDB), and after that the writes are emitted into the graph (and processed by the bases). If we only write to the WAL here, then isn't there a window between when we ACK and when we write to the WAL here where we can crash and lose ACKed writes?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+
+    fn lookup(&self, columns: &[usize], key: &KeyType) -> LookupResult {
+        let key_values = match *key {
+            KeyType::Single(a) => vec![a],
+            KeyType::Double(ref r) => vec![&r.0, &r.1],
+            KeyType::Tri(ref r) => vec![&r.0, &r.1, &r.2],
+            KeyType::Quad(ref r) => vec![&r.0, &r.1, &r.2, &r.3],
+            KeyType::Quin(ref r) => vec![&r.0, &r.1, &r.2, &r.3, &r.4],
+            KeyType::Sex(ref r) => vec![&r.0, &r.1, &r.2, &r.3, &r.4, &r.5],
+        };
```
jonhoo: These `vec!`s are a little sad. We can't get around it by making a chained iterator or something (which doesn't allocate)? I *think* below we only ever read the keys left to right, and only ever needs refs...

jonhoo: Or better yet, make `serialize_prefix` take a `&KeyType` as its second argument.

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+                let index_id = self.get_index_id(columns)
+                    .expect("lookup on non-indexed column set");
+                let prefix = Self::serialize_prefix(index_id, &key_values);
+                let values = db.prefix_iterator(&prefix);
+                if index_id > 0 {
+                    // For non-primary indices we need to first retrieve the primary index key
+                    // through our secondary indices, and then call `.get` again on that.
+                    values
+                        .map(|(_key, value)| {
+                            let raw_row = db.get(&value)
```
jonhoo: Is there some kind of `multi_get` we could use here? Might be a lot faster...

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+    }
+
+    fn cloned_records(&self) -> Vec<Vec<DataType>> {
+        self.all_rows()
+            .map(|(_, ref value)| bincode::deserialize(&value).unwrap())
+            .collect()
+    }
+
+    fn rows(&self) -> usize {
+        self.all_rows().count()
```
jonhoo: This seems *really* bad! Can we do any better? Is there a RocksDB "count number of values for a prefix" operation? Alternatively, can we keep the row count around somewhere? It seems *really* surprising for the innocuous-sounding `state.rows()` to **scan the entire base store**.

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+    }
+
+    fn clear(&mut self) {
+        // Would potentially be faster to just drop self.db and call DB::Destroy:
```
jonhoo: Yeah, that sounds better. When is this method every called though?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+            // that we create a directory for each base shard:
+            opts.set_wal_dir(path.join(&name));
+        }
+
+        // Create prefixes by Self::transform_fn on all new inserted keys:
+        let transform = SliceTransform::create("key", Self::transform_fn, Some(Self::in_domain_fn));
+        opts.set_prefix_extractor(transform);
+
+        // Assigns the number of threads for RocksDB's low priority background pool:
+        opts.increase_parallelism(params.persistence_threads);
```
jonhoo: Are you sure these workers end up being shared among all base nodes?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+
+        // Use a hash linked list since we're doing prefix seeks.
+        opts.set_allow_concurrent_memtable_write(false);
+        opts.set_memtable_factory(rocksdb::MemtableFactory::HashLinkList {
+            bucket_count: 1_000_000,
```
jonhoo: Is this never resized, even if it becomes full?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+            // that was added first. This means that we can't consider the keys unique though, so
+            // we'll append a sequence number.
+            let pk_index = &self.indices[0];
+            let pk = pk_index.columns.iter().map(|i| &r[*i]).collect::<Vec<_>>();
+            (&self.indices[1..], self.serialize_key(0, &pk, pk_index.seq))
+        };
+
+        // First insert the actual value for our primary index:
+        let pk_value = bincode::serialize(&r).unwrap();
+        batch.put(&serialized_pk, &pk_value).unwrap();
```
jonhoo: I'm a little surprised that the `rocksdb` don't have a way of sending a value without storing it in an intermediate `&[u8]` buffer... Oh well.

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+            Some(ref pk) => &pk[..] == columns,
+            _ => false,
+        };
+
+        if existing || is_pk {
+            return;
+        }
+
+        let cols = Vec::from(columns);
+        let index_id = self.indices.len() as u32;
```
jonhoo: `+1` if `self.primary_key.is_some()`?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+
+    fn get_index_id(&self, columns: &[usize]) -> Option<IndexID> {
+        self.indices
+            .iter()
+            .position(|index| &index.columns[..] == columns)
+            .and_then(|id| {
+                let id = id as IndexID;
+                // Increment the index ID with one if we have a primary key,
+                // as that'll always be index 0:
+                Some(match self.primary_key {
```
jonhoo: Minor, but I'd do `if self.primary_key.is_some()`

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+    // > db.NewIterator(). For prefix filtering to work properly,
+    // > "prefix_extractor" and "comparator" must be such that the following
+    // > properties hold:
+    //
+    // > 1) key.starts_with(prefix(key))
+    // > 2) Compare(prefix(key), key) <= 0.
+    // > 3) If Compare(k1, k2) <= 0, then Compare(prefix(k1), prefix(k2)) <= 0
+    // > 4) prefix(prefix(key)) == prefix(key)
+    //
+    // NOTE(ekmartin): Encoding the key size in the key increases the total size with 4 bytes.
```
jonhoo: 8 bytes?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+            return Vec::from(key);
+        }
+
+        // IndexID is a u32, so bincode uses 4 bytes to serialize it (which we'll skip past):
+        let start = 4;
+        // We encoded the size of the key itself with a u64, which bincode uses 8 bytes to encode:
+        let size_offset = start + 8;
+        let key_size: u64 = bincode::deserialize(&key[start..size_offset]).unwrap();
+        let prefix_len = size_offset + key_size as usize;
+        let mut bytes = Vec::from(key);
```
jonhoo: Can't you get rid of the below and just make this `Vec::from(&key[..prefix_len])`?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+    // up here in transform_fn. Example:
+    // [DataType::Int(1), DataType::BigInt(10)] would be serialized as:
+    // 2u64 (vec length), 0u32 (enum variant), 1i32 (value), 1u32 (enum variant), 1i64 (value)
+    // By stepping through the serialized bytes and checking each enum variant we would know
+    // when we reached the end, and could then with certainty say whether we'd already
+    // prefix transformed this key before or not
+    // (without including the byte size of Vec<DataType>).
+    fn transform_fn(key: &[u8]) -> Vec<u8> {
+        if key == META_KEY {
+            return Vec::from(key);
```
jonhoo: `unreachable!()` given `in_domain_fn`?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+    // Decides which keys the prefix transform should apply to.
+    fn in_domain_fn(key: &[u8]) -> bool {
+        key != META_KEY
+    }
+
+    // A key is built up of five components:
+    //  * `index_id`
+    //     Uniquely identifies the index in self.indices.
+    //  * `row_size`
+    //      The byte size of `row` when serialized with binode.
```
jonhoo: Maybe call these `key_` instead of `row_` given that they don't actually hold "rows".

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+    // TODO(ekmartin): This will put exactly the values that are given, and can only be retrieved
+    // with exactly those values. I think the regular state implementation supports inserting
+    // something like an Int and retrieving with a BigInt.
+    fn insert(&mut self, batch: &mut WriteBatch, r: Vec<DataType>) {
+        for index in self.indices.iter_mut() {
+            index.seq += 1;
+        }
+
+        let (secondary_indices, serialized_pk) = if let Some(ref pk_cols) = self.primary_key {
+            let pk = pk_cols.iter().map(|i| &r[*i]).collect::<Vec<_>>();
```
jonhoo: I *think* that if `serialize_prefix` took a `KeyType`, you wouldn't need to `collect()` (and allocate) here.

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+            (&self.indices[..], Self::serialize_prefix(0, &pk))
+        } else {
+            // For bases without primary keys we store the actual row values keyed by the index
+            // that was added first. This means that we can't consider the keys unique though, so
+            // we'll append a sequence number.
+            let pk_index = &self.indices[0];
+            let pk = pk_index.columns.iter().map(|i| &r[*i]).collect::<Vec<_>>();
```
jonhoo: Same re: `KeyType` and `self.serialize_key`

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+            (&self.indices[1..], self.serialize_key(0, &pk, pk_index.seq))
+        };
+
+        // First insert the actual value for our primary index:
+        let pk_value = bincode::serialize(&r).unwrap();
+        batch.put(&serialized_pk, &pk_value).unwrap();
+
+        // Then insert primary key pointers for all the secondary indices:
+        for (i, index) in secondary_indices.iter().enumerate() {
```
jonhoo: Won't the `i` here be off is we don't have a primary key, have multiple non-primary indices, and thus end up returning `&self.indices[1..]` above?

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+            // Construct a key with the index values, and serialize it with bincode:
+            let key = index.columns.iter().map(|i| &r[*i]).collect::<Vec<_>>();
```
jonhoo: Again with `KeyType` to `serialize_key`

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+        let db = self.db.as_ref().unwrap();
+        let mut delete_cols: Vec<_> = self.indices
+            .iter()
+            .enumerate()
+            .map(|(i, index)| {
+                if self.primary_key.is_some() {
+                    (i + 1, &index.columns)
+                } else {
+                    (i, &index.columns)
+                }
```
jonhoo: Wouldn't it be easier to just store the `IndexId` directly with the indices instead so we wouldn't need to remember to do this mapping everywhere? Or alternatively store the primary key *as* the first index and just have a `first_is_unique: bool`.

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+            })
+            .collect();
```
jonhoo: Doing this `collect` on *every* `remove` seems pretty unfortunate...

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+
+        if let Some(ref pk_cols) = self.primary_key {
+            delete_cols.push((0, pk_cols));
+        }
+
+        for (i, columns) in delete_cols.into_iter() {
+            let index_row = columns.iter().map(|i| &r[*i]).collect::<Vec<_>>();
+            let serialized_key = Self::serialize_prefix(i as u32, &index_row);
+            for (key, _value) in db.prefix_iterator(&serialized_key) {
+                batch.delete(&key).unwrap();
```
jonhoo: This doesn't seem right. `remove` should remove *one*, not *all* rows. And specifically should only remove one row that *exactly matches* `r`, not just where the key is the same.. And I suspect that means you have to remove one row from the first index first, and then search the secondary indices for the entry that references the removed row from the first index.

core/src/local/persistent_state.rs
ekmartin: #63 moved the write-ACKs to `process.rs` after `materialize` has been called, so `process_records` should finish before that happens - right?

jonhoo: Ohhh, yeah, you're right. My bad.

core/src/local/persistent_state.rs
ekmartin: Nice catch! Fixed in a3e9e1fbccd2bd4f5cacb5aa45e494d41d63cb78.

core/src/local/persistent_state.rs
ekmartin: Good point, moved the primary key to `self.indices` now.

core/src/local/persistent_state.rs
ekmartin: Yeah, good idea. Changed `serialize_key` and `serialize_prefix` to take a `KeyType` - definitely nicer.

core/src/local/persistent_state.rs
ekmartin: One would think so, but it seems like RocksDB always uses slice transforms to create keys when you use a hash-based memtable factory (regardless of whether the domain function matches or not). The domain function then only applies to whether the slice transform is used to create bloom filters or not. 

Sort of described [here](https://github.com/facebook/rocksdb/blob/master/include/rocksdb/slice_transform.h#L42-L57), but still not too clear.

Added a comment about it in the `transform_fn` at least.

core/src/local/persistent_state.rs
ekmartin: Nice, thanks :)

core/src/local/persistent_state.rs
ekmartin: Good point!

core/src/local/persistent_state.rs
ekmartin: There is, but it doesn't help with performance:

> Q: If I want to retrieve 10 keys from rocksdb, is it better to batch them and use MultiGet() versus issuing 10 individual Get() calls?
> A: The performance is similar. MultiGet() reads from the same consistent view, but it is not faster.

Might be cleaner regardless though, as it gives us one less FFI call. `rust-rocksdb` doesn't support it yet, but shouldn't be too hard to add - I'll look into it.

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,891 @@
...
+        let mut batch = None;
+
+        // Build the new index for existing values:
+        self.all_rows().for_each(|(ref pk, ref value)| {
+            seq += 1;
+            let row: Vec<DataType> = bincode::deserialize(&value).unwrap();
+            let index_key = columns.iter().map(|i| &row[*i]).collect::<Vec<_>>();
+            let key = self.serialize_key(index_id, &index_key, seq);
+            let b = batch.get_or_insert_with(|| WriteBatch::default());
+            b.put(&key, &pk).unwrap();
```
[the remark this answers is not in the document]
ekmartin: Good point - I added a `.chunks()` call to make sure that it only builds write batches containing a sane amount of rows (currently set to 512), which are then synced to disk per chunk.

core/src/local/persistent_state.rs
ekmartin: The only caller was the `unseed()` function in `MockGraph` so I removed it. Less code 👍

core/src/local/persistent_state.rs
ekmartin: Yup! See https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide#sharing-cache-and-thread-pool

(also, `vote` with a lot of shards and a single persistence thread doesn't create one thread per shard)

core/src/local/persistent_state.rs
jonhoo: Ah, I guess I was (erroneously) thinking of RocksDB as being external, and that you'd have to wait for an RTT or something, but that's obviously not true. FFI calls seem fine.

core/src/local/persistent_state.rs
jonhoo: Yeah, the concern I specifically have is that hash maps stray further and further from `O(1)` the larger the buckets grow. If they never resize, then we could end up in pretty sad territory.

core/src/local/persistent_state.rs
```diff
@@ -382,19 +370,19 @@ impl PersistentState {
...
+    //  * `key`
     //      The actual index values.
     //  * `epoch`
     //      Incremented on each PersistentState initialization.
     //  * `seq`
     //      Sequence number since last epoch.
-    fn serialize_key(&self, index_id: IndexID, row: &Vec<&DataType>, seq: IndexSeq) -> Vec<u8> {
-        let size: u64 = bincode::serialized_size(&row).unwrap();
-        bincode::serialize(&(index_id, size, row, self.epoch, seq)).unwrap()
+    fn serialize_key(&self, index_id: IndexID, key: &KeyType, seq: IndexSeq) -> Vec<u8> {
```
jonhoo: Hmm.. I think we'd rather serialize the keys the way we used to (i.e., without the `KeyType` enum variant field). It'd basically mean matching on `KeyType` here and then serializing the elements individually. Don't know how important saving 8 bytes on every key is though..

core/src/local/persistent_state.rs
```diff
@@ -153,20 +157,19 @@ impl State for PersistentState {
...
         // Build the new index for existing values:
-        self.all_rows().for_each(|(ref pk, ref value)| {
-            seq += 1;
-            let row: Vec<DataType> = bincode::deserialize(&value).unwrap();
-            let index_key = KeyType::from(columns.iter().map(|i| &row[*i]));
-            let key = self.serialize_key(index_id, &index_key, seq);
-            let b = batch.get_or_insert_with(|| WriteBatch::default());
-            b.put(&key, &pk).unwrap();
-        });
+        for chunk in self.all_rows().chunks(INDEX_BATCH_SIZE).into_iter() {
```
jonhoo: I don't think you need to call `into_iter()` explicitly here?

core/src/local/persistent_state.rs
```diff
@@ -442,29 +442,56 @@ impl PersistentState {
...
+        // First retrieve the actual stored row that matches r:
         if let Some(ref pk_cols) = self.primary_key {
-            delete_cols.push((0, pk_cols));
-        }
+            let pk = KeyType::from(pk_cols.iter().map(|i| &r[*i]));
+            let prefix = Self::serialize_prefix(0, &pk);
+            let raw_value = db.get(&prefix).unwrap();
+            if let Some(raw) = raw_value {
+                let value: Vec<DataType> = bincode::deserialize(&*raw).unwrap();
+                if r != &value[..] {
```
jonhoo: I think maybe this should `assert_eq!(r, &value[..])`, since we never expect get a negative that doesn't match a stored row.

ekmartin: Good point. Added an `.expect()` for the case where the PK row isn't found as well. Not sure if there should be something similar for non-PK removals, e.g. where it can't find at least one row?

core/src/local/persistent_state.rs
```diff
@@ -410,28 +406,29 @@ impl PersistentState {
     // with exactly those values. I think the regular state implementation supports inserting
     // something like an Int and retrieving with a BigInt.
     fn insert(&mut self, batch: &mut WriteBatch, r: &[DataType]) {
-        for index in self.indices.iter_mut() {
+        for index in self.indices[1..].iter_mut() {
```
jonhoo: Don't you need to increment `seq` for the first index if it's not a primary key too?

core/src/local/persistent_state.rs
```diff
@@ -410,28 +406,29 @@ impl PersistentState {
...
-            (&self.indices[1..], self.serialize_key(0, &pk, pk_index.seq))
+        let serialized_pk = {
+            let pk = KeyType::from(self.indices[0].columns.iter().map(|i| &r[*i]));
+            if self.has_unique_index {
+                Self::serialize_prefix(0, &pk)
+            } else {
+                // For bases without primary keys we store the actual row values keyed by the index
+                // that was added first. This means that we can't consider the keys unique though, so
+                // we'll append a sequence number.
+                self.indices[0].seq += 1;
```
jonhoo: Given the loop further up that adds 1 to all `seq`s, why is it necessary to add again here?

ekmartin: > Don't you need to increment seq for the first index if it's not a primary key too?

That happens here. Might be more explicit to have a conditional in the for loop though, to avoid spreading out the logic.

core/src/local/persistent_state.rs
```diff
@@ -442,29 +442,56 @@ impl PersistentState {
...
+        } else {
+            let pk_index = &self.indices[0];
+            let pk = KeyType::from(pk_index.columns.iter().map(|i| &r[*i]));
+            let prefix = Self::serialize_prefix(0, &pk);
+            let key = db.prefix_iterator(&prefix).find(|(_, raw_value)| {
+                let value: Vec<DataType> = bincode::deserialize(&*raw_value).unwrap();
+                r == &value[..]
+            });
+
+            if key.is_none() {
```
jonhoo: Probably `assert` here too

core/src/local/persistent_state.rs
jonhoo: In general, every removal should remove exactly one row. If they don't, it means we produced an incorrect negative, which should never happen.

core/src/local/persistent_state.rs
jonhoo: Ohhh, I missed the `[0]` part. My bad.

dataflow/src/ops/latest.rs
ekmartin: Fixed here: https://github.com/mit-pdos/distributary/pull/72/commits/3a4a7c235f1e17dc8ec1084785be25a37e539a7c#diff-4822603b49efd860fa34b9ae63ccf3bdR99

dataflow/src/ops/grouped/mod.rs
ekmartin: Yeah, agreed. Tried briefly but got shot down by the borrow checker, so added a TODO for now.

dataflow/src/domain/mod.rs
ekmartin: It ends up cloning the `Row` without it, probably because `Row` now derives `Clone` as well (which is needed by `Cow`). I changed it to double deref `row` further down instead of the variable though, seemed cleaner. 

https://github.com/mit-pdos/distributary/pull/72/files#diff-f3207323d6f34ad3c124fbcb4cb87a40R1477

core/src/local/persistent_state.rs
ekmartin: Changed this to use an estimated count property from RocksDB now.

core/src/local/persistent_state.rs
```diff
@@ -0,0 +1,1065 @@
...
+// Store index information in RocksDB to avoid rebuilding indices on recovery.
+#[derive(Default, Serialize, Deserialize)]
+struct PersistentMeta {
+    indices: Vec<Vec<usize>>,
+    epoch: IndexEpoch,
+}
+
+#[derive(Clone)]
+struct PersistentIndex {
+    column_family: ColumnFamily,
```
ekmartin: @jonhoo: had to add `impl unsafe Send` to [`ColumnFamily`](https://github.com/spacejam/rust-rocksdb/blob/701669d0763a0fb30708d36b1af2e152bda1e63d/src/lib.rs#L203-L208) in rust-rocksdb for this to work, but I think it should be fine as rust-rocksdb already stores `ColumnFamily` objects in a `BTreeMap` [internally](https://github.com/spacejam/rust-rocksdb/blob/701669d0763a0fb30708d36b1af2e152bda1e63d/src/lib.rs#L81), and the `DB` object where they're stored in is `unsafe Send` itself.

The alternative would be to just rely on rust-rocksdb's [`cf_handle`](https://github.com/spacejam/rust-rocksdb/blob/701669d0763a0fb30708d36b1af2e152bda1e63d/src/db.rs#L881-L884), which clones the `*mut ffi::rocksdb_column_family_handle_t` pointer. It takes a `&str` though, but could just keep something like a `vec!["0", 1", "2", ... to self.indices.len()]` around on `PersistentState` to avoid having to allocate through `to_string()` all the time.

core/src/local/persistent_state.rs
```diff
@@ -444,70 +451,77 @@ impl PersistentState {
...
-                self.serialize_key(0, &pk, self.indices[0].seq)
+                self.serialize_key(&pk, self.indices[0].seq)
             }
         };
 
         // First insert the actual value for our primary index:
         let pk_value = bincode::serialize(&r).unwrap();
-        batch.put(&serialized_pk, &pk_value).unwrap();
+        let db = self.db.as_ref().unwrap();
+        let value_cf = db.cf_handle("0").unwrap();
```
jonhoo: I assume this operation is relatively cheap?

core/src/local/persistent_state.rs
```diff
@@ -444,70 +451,77 @@ impl PersistentState {
...
-            batch.delete(&primary_key).unwrap();
+        let value_cf = db.cf_handle("0").unwrap();
+        let mut do_remove = move |primary_key: &[u8]| {
+            // Delete the value row first (primary index):
+            batch.delete_cf(value_cf, &primary_key).unwrap();
 
             // Then delete any references that point _exactly_ to
             // that row (i.e. not all matching rows):
-            for (i, index) in indices.iter().enumerate() {
+            for (i, index) in self.indices[1..].iter().enumerate() {
```
jonhoo: This might read better as:
```rust
for (i, index) in self.indices.iter().enumerate().skip(1) {
```
So that you can get rid of the `index_id` computation below.

core/src/local/persistent_state.rs
jonhoo: Yeah, no, that seems fine then.

core/src/local/persistent_state.rs
jonhoo: Haha, never mind then given the next change :p

core/src/local/persistent_state.rs
```diff
@@ -279,6 +297,9 @@ impl PersistentState {
         });
 
         let full_name = format!("{}.db", name);
+        // We use a column for each index, and one for meta information.
+        // When opening the DB the exact same column families needs to be used,
+        // so we'll have to retrieve the existing ones first:
         let column_family_names = match DB::list_cf(&opts, &full_name) {
             Ok(cfs) => cfs,
```
jonhoo: Couldn't we do the thing you do in `add_key` here if `cfs.len()` != meta.indices.len()`?

PR #75, opened 2018-04-19
Evict from readers

jonhoo:


dataflow/Cargo.toml
```diff
@@ -5,9 +5,10 @@ authors = ["The Distributary Developers"]
 
 [dependencies]
 buf_redux = "0.6.1"
-evmap = "2.0.2"
+evmap = { git = "https://github.com/ms705/rust-evmap" }
```
jonhoo: I assume this is to get `evmap` with `rahashmap` so we can evict random keys?

ms705: Yes, though we can perhaps merge those changes into upstream `evmap` now that rahashmap has been updated. I'll make a PR.

dataflow/src/backlog/mod.rs
```diff
@@ -221,7 +226,12 @@ impl WriteHandle {
     where
         I: IntoIterator<Item = Record>,
     {
-        self.handle.add(&self.key[..], self.cols, rs)
+        let mem_delta = self.handle.add(&self.key[..], self.cols, rs);
+        if mem_delta > 0 {
+            self.mem_size += mem_delta as usize;
+        } else if mem_delta < 0 {
+            self.mem_size = self.mem_size.checked_sub(mem_delta as usize).unwrap();
```
jonhoo: This doesn't seem right? `mem_delta as usize` when `mem_delta < 0`?

dataflow/src/backlog/mod.rs
```diff
@@ -231,6 +241,37 @@ impl WriteHandle {
     pub(crate) fn is_partial(&self) -> bool {
         self.partial
     }
+
+    /// Evict `count` randomly selected keys from state and return them along with the number of
+    /// bytes that will be freed once the underlying `evmap` applies the operation.
+    pub fn evict_random_key(&mut self, rng: &mut ThreadRng) -> u64 {
+        let mut bytes_to_be_freed = 0;
+        if self.mem_size > 0 {
+            match self.handle.empty_at_index(rng.gen()) {
```
jonhoo: Shouldn't this generate something that's in the range of indices? Or is it automatically wrapped around so it always hits a bucket?

dataflow/src/backlog/mod.rs
```diff
@@ -231,6 +241,37 @@ impl WriteHandle {
...
+                None => (),
+                Some(vs) => {
+                    let size: u64 = vs.into_iter().map(|r| r.deep_size_of() as u64).sum();
+                    bytes_to_be_freed += size;
```
jonhoo: So this doesn't take into account the fact that `evmap` stores the index twice?

dataflow/src/backlog/mod.rs
```diff
@@ -231,6 +241,37 @@ impl WriteHandle {
...
+
+impl SizeOf for WriteHandle {
+    fn size_of(&self) -> u64 {
+        use std::mem::size_of;
+
+        size_of::<Self>() as u64
+    }
+
+    fn deep_size_of(&self) -> u64 {
+        self.mem_size as u64
```
jonhoo: Should this not include the size of `Self`?

dataflow/src/backlog/multiw.rs
```diff
@@ -26,6 +27,16 @@ impl Handle {
         }
     }
 
+    /// Evict `count` randomly selected keys from state and return them along with the number of
+    /// bytes freed.
+    pub fn empty_at_index(&mut self, index: usize) -> Option<&Vec<Vec<DataType>>> {
```
jonhoo: I think this might see a stale version of the data at a given index. Remember how some writes aren't applied yet to the handle we receive after swapping?

dataflow/src/backlog/mod.rs
ms705: It's a subtraction of the absolute value, so still correct?

dataflow/src/backlog/mod.rs
ms705: I believe rahashmap automatically wraps. @fintelia ?

dataflow/src/backlog/mod.rs
ms705: The rows are deduplicated, and for simplicity we only count memory in rows, not the index entries. Coudl change this, but it wouldn't change any of the fundamentals.

dataflow/src/domain/mod.rs
```diff
@@ -1389,11 +1398,20 @@ impl Domain {
...
+                                    // We are a reader, which has its own kind of state
+                                    let mut size = 0;
+                                    n.with_reader(|r| size = r.state_size().unwrap_or(0))
+                                        .unwrap();
+                                    size
+                                } else {
+                                    // Not a reader, state is with domain
+                                    self.state
+                                        .get(&local_index)
+                                        .filter(|state| state.is_partial())
```
jonhoo: So reported memory use is only the size of partial states then?

dataflow/src/backlog/mod.rs
ms705: Sure, can include it, but it's just a constant 16 (?) bytes.

dataflow/src/domain/mod.rs
jonhoo: I don't see us making the same distinction from `Reader`s, even though presumably the same should apply to them?

ms705: `UpdateStateSize` is only used for eviction, and there's no point in evicting from anything other than a partial state.

`GetStatistics` includes all states.

dataflow/src/domain/mod.rs
```diff
@@ -1915,7 +1933,18 @@ impl Domain {
...
+
                                         for key in backfill_keys.iter() {
+                                            let size = wh.with_key(&key[..])
+                                                .try_find_and(|rs| {
+                                                    rs.iter().map(|r| r.deep_size_of()).sum()
+                                                })
+                                                .map(|r| r.0.unwrap_or(0))
+                                                .unwrap_or(0);
+                                            wh.mem_size =
+                                                wh.mem_size.checked_sub(size as usize).unwrap();
```
jonhoo: `mark_filled` *should* only be called if the current state is empty, so I think this should (if anything) just assert that the state for that key is 0. Or am I missing something?

dataflow/src/domain/mod.rs
ms705: Good point re readers, will fix.

dataflow/src/domain/mod.rs
```diff
@@ -1974,6 +2003,18 @@ impl Domain {
...
+
+                                                let size = wh.with_key(&miss[..])
+                                                    .try_find_and(|rs| {
+                                                        rs.iter().map(|r| r.deep_size_of()).sum()
+                                                    })
+                                                    .map(|r| r.0.unwrap_or(0))
+                                                    .unwrap_or(0);
+
+                                                wh.mem_size =
+                                                    wh.mem_size.checked_sub(size as usize).unwrap();
```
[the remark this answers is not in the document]
ms705: This is the thing I mentioned when we talked. It can't be done in `mark_hole` because we do not have access to the `WriteHandle`'s `self.mem_size` there.

dataflow/src/domain/mod.rs
```diff
@@ -2453,44 +2494,63 @@ impl Domain {
...
 
-                        let (key_columns, keys, bytes) = {
-                            let k = self.state[&node].evict_random_keys(100);
-                            (k.0.to_vec(), k.1, k.2)
-                        };
-                        freed += bytes;
+                        if self.nodes[&node].borrow().is_reader() {
+                            // we can only evict one key a time here because the freed memory
+                            // calculation is based on the key that *will* be evicted. We may count
+                            // the same individual key twice if we batch evictions here.
```
jonhoo: I'm not sure I understand this logic? What is stopping us from clearing multiple buckets in the `Reader` at one time?

dataflow/src/node/special/reader.rs
```diff
@@ -130,6 +135,20 @@ impl Reader {
...
+    /// Evict a randomly selected key, returning the number of bytes evicted.
+    /// Note that due to how `evmap` applies the evictions asynchronously, we can only evict a
+    /// single key at a time here.
+    pub fn evict_random_key(&mut self) -> u64 {
+        let mut bytes_freed = 0;
+        if let Some(ref mut handle) = self.writer {
+            use rand;
+            let mut rng = rand::thread_rng();
+            bytes_freed = handle.evict_random_key(&mut rng);
+            handle.swap();
```
jonhoo: Why is it necessary to `swap` here?

dataflow/src/domain/mod.rs
ms705: See the comment. We may choose an index that maps to the same bucket, and double-count it.

dataflow/src/backlog/mod.rs
jonhoo: My question was more what happens if you cast a negative `isize` into ` usize`.

dataflow/src/domain/mod.rs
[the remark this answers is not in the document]
jonhoo: Hmm... That seems fixable. Just make https://github.com/mit-pdos/distributary/blob/master/dataflow/src/backlog/mod.rs#L120 be an `&'a mut WriteHandle` instead?

dataflow/src/domain/mod.rs
jonhoo: If `WriteHandle` had a `evict_random_keys`, it could just make sure it doesn't do that in the loop in https://github.com/mit-pdos/distributary/pull/75/files#diff-117bbb5259149102b8211b4567837f17R250 no?

dataflow/src/domain/mod.rs
ms705: 27cb886dfd3fba9245ca3e3bed231816480db369

dataflow/src/domain/mod.rs
ms705: `mark_filled` calls `self.handle.clear()`, so at least potentially throws away data. I can add an assertion there, but then I don't see what the purpose of `mark_filled` is at all -- it would have nothing to do?

jonhoo: You can think of `mark_filled` as removing the thunk that indicates that there's a hole. What it *actually* does is insert `Vec::new` for the given key so that we know that it isn't missing even if the backfill produced no rows for that key. In all cases though, the current state for that key *should* be empty. Even if it does do `clear` though, can't you do the same thing as for eviction and sum up the size of the cleared rows?

ms705: That's what I'm doing here. I cannot do it inside `mark_filled` due to lack of access to `self.mem_size`, and I have to do it as a lookup because `evmap::WriteHandle::clear` does not return anything as it just enqueues an op. We could change `evmap` to do an internal lookup and return the cleared rows and/or the size, I guess.

jonhoo: Ah, i see what you mean. Yes, you would have to do a lookup before clearing (though `mark_filled` will have access to `mem_size` with the change below, so this code could move in there). That said, I think what I'd really do is simply assert that there are no rows for that key (like we do in local state: https://github.com/mit-pdos/distributary/blob/aebefadc1fa540c0a363a5c53cf418b17efeed8e/core/src/local/single_state.rs#L241).

dataflow/src/backlog/mod.rs
```diff
@@ -122,21 +122,36 @@ pub(crate) struct WriteHandle {
...
+    handle: &'a WriteHandle,
     key: Key<'a>,
 }
 
 impl<'a> MutWriteHandleEntry<'a> {
     pub fn mark_filled(self) {
-        self.handle.clear(self.key)
+        if let Some((None, _)) = self.handle
+            .handle
+            .meta_get_and(self.key.clone(), |rs| rs.is_empty())
```
ms705: This clone is unfortunate, but I think required, as we need self.key later on line 139.

jonhoo: I think you can just use `&*self.key` here?

dataflow/src/backlog/mod.rs
```diff
@@ -122,21 +122,36 @@ pub(crate) struct WriteHandle {
...
+        } else {
+            unreachable!("attempted to fill already-filled key");
+        }
     }
 
     pub fn mark_hole(self) {
-        self.handle.empty(self.key)
+        let size = self.handle
+            .handle
+            .meta_get_and(self.key.clone(), |rs| {
```
jonhoo: Same here; just `&*self.key`

PR #76, opened 2018-04-25
Refactor LookupResult

ekmartin:
Implements a `Cow`-like enum where the `Owned` branch is a `Vec<Vec<DataType>>` and `Borrowed` is a `[Row]` - this lets us avoid `Row`'s `Rc` wrapper in the `Owned` case.

PR #77, opened 2018-04-25
Implement query_through for Project

ekmartin:


dataflow/src/ops/project.rs
```diff
@@ -122,6 +124,79 @@ impl Ingredient for Project {
...
+                                new_r.append(&mut a.clone());
+                            }
+
+                            Cow::from(new_r)
+                        })) as Box<_>,
+                        None => Box::new(rs.into_iter().map(|r| Cow::from(r.unpack()))) as Box<_>,
+                    };
+
+                    Some(Some(r))
+                }
```
jonhoo: The repetition here is pretty sad :/ Will this get better with the `into_iter` from #76?

dataflow/src/node/process.rs
```diff
@@ -202,6 +202,40 @@ impl Node {
...
+                                .find(|n| n.borrow().global_addr() == parent)
+                                .unwrap();
+                            miss.on = *parent_node.borrow().local_addr();
+                            *col = parent_col;
+                        }
+                    }
+                };
+
+                for miss in misses.iter_mut() {
+                    if miss.on == addr {
```
jonhoo: Why is this condition here? I don't think there are any operators that do lookups on themselves and are *also* query through? At least for now.

dataflow/src/node/process.rs
```diff
@@ -202,6 +202,40 @@ impl Node {
                     _ => {}
                 };
 
+                // When we miss at a can_query_through node we have to replay from
+                // its parent instead of that node, so we'll update the misses:
+                let reroute_miss = |internal: &NodeOperator, miss: &mut Miss| {
```
jonhoo: I think this has to recurse. Consider a join over a filtered projection. There are two layers of `query_through`.

dataflow/src/node/process.rs
```diff
@@ -287,6 +259,34 @@ impl Node {
...
+                "can't query_through with more than one parent"
+            );
+
+            let (parent, parent_col) = parents[0];
+            let parent_node = nodes
+                .values()
+                .find(|n| n.borrow().global_addr() == parent)
+                .unwrap();
+
+            miss.on = *parent_node.borrow().local_addr();
```
jonhoo: Why is this *inside* the `for` loop? I'd put it outside and assert that all the columns you resolve resolve to the same parent node.

dataflow/src/ops/filter.rs
```diff
@@ -170,47 +170,49 @@ impl Ingredient for Filter {
...
-                                    Operator::Equal => d == v,
-                                    Operator::NotEqual => d != v,
-                                    Operator::Greater => d > v,
-                                    Operator::GreaterOrEqual => d >= v,
-                                    Operator::Less => d < v,
-                                    Operator::LessOrEqual => d <= v,
-                                    _ => unimplemented!(),
+        self.lookup(*self.src, columns, key, domains, states)
+            .and_then(|result| {
+                let f = self.filter.clone();
```
jonhoo: I don't understand why `f` here has to `clone`. Not introduced by your change though..

dataflow/src/processing.rs
```diff
@@ -218,7 +219,7 @@ where
         parent: LocalNodeIndex,
         columns: &[usize],
         key: &KeyType,
-        domain: &DomainNodes,
+        domains: &DomainNodes,
```
jonhoo: `nodes`. `DomainNodes` does not contain multiple domains.

jonhoo:
I think that for this to work we also have to change the other operators that do lookups to use `self.lookup`, right? A quick search yields [`grouped`](https://github.com/mit-pdos/distributary/blob/caaa1a62eb1972cac859dd6926ff892980cd98d8/dataflow/src/ops/grouped/mod.rs#L178), [`latest`](https://github.com/mit-pdos/distributary/blob/caaa1a62eb1972cac859dd6926ff892980cd98d8/dataflow/src/ops/latest.rs#L60), [`trigger`](https://github.com/mit-pdos/distributary/blob/caaa1a62eb1972cac859dd6926ff892980cd98d8/dataflow/src/ops/trigger.rs#L127), and [`topk`](https://github.com/mit-pdos/distributary/blob/caaa1a62eb1972cac859dd6926ff892980cd98d8/dataflow/src/ops/topk.rs#L146). `join` seems to already use `self.join`, and should be all good.

PR #79, opened 2018-04-27
Don't use timestamps in secondary keys

ekmartin:
Secondary keys are now suffixed with the primary key instead. This makes `.remove()` much simpler, as we can delete rows for all indices without any reads (except the initial `.lookup` in `Base`).


jonhoo:
LGTM :tada:

PR #80, opened 2018-05-17
Distinct

jsegaran:
This is the distinct operator, but at the moment it requires full materialization to work at the moment, partial isn't yet supported for this operator.


dataflow/src/ops/distinct.rs
```diff
@@ -0,0 +1,252 @@
+use std::collections::HashMap;
+
+use prelude::*;
+
+/// This will get Distinct elements from each group
+/// TODO
+
```
jonhoo: It' be great if you could expand this comment to explain what the operator does, how, and what it's limitations are :) In Rust, you also have to not leave a space between a documentation comment and the item it is documenting (i.e., the newline above has to go away).

dataflow/src/ops/distinct.rs
```diff
@@ -0,0 +1,252 @@
...
+pub struct Distinct {
+    // Parent Node
+    src: IndexPair,
+
+    us: Option<IndexPair>,
+
+    group_by: Vec<usize>,
+
+    // A list of records that have been seen
+    seen: HashMap<Vec<DataType>, bool>,
```
jonhoo: It doesn't look like this field is currently used anywhere? Was the idea to use this instead of the
```
let mut seen = HashMap::new();
```
inside `on_input`?

dataflow/src/ops/distinct.rs
```diff
@@ -0,0 +1,252 @@
...
+}
+
+impl Distinct {
+    pub fn new(
+        src: NodeIndex,
+        group_by: Vec<usize>,
+        ) -> Self {
+
+        // TODO add implementation for group by
```
jonhoo: Not sure what this `TODO` refers to?

dataflow/src/ops/distinct.rs
```diff
@@ -0,0 +1,252 @@
...
+}
+
+impl Ingredient for Distinct {
+    /// Returns a clone of the node
+    fn take(&mut self) -> NodeOperator {
+        Clone::clone(self).into()
+    }
+
+    fn ancestors(&self) -> Vec<NodeIndex> {
+        // I think this is right?
```
jonhoo: Yup, it is!

dataflow/src/ops/distinct.rs
jonhoo: It's also likely a good idea to document that these do not actually hold *all* records that have been seen, just the ones that have been seen in *this* packet. That is, they are cleared for every input.

dataflow/src/ops/distinct.rs
```diff
@@ -0,0 +1,252 @@
...
+            if seen.contains_key(&group){
+                if !positive{
+                    seen.remove(&group);
+                    cleaned.push(rec.clone());
+                }
+            } else {
+                cleaned.push(rec.clone());
+                seen.insert(group.clone(), true);
+            }
+        }
```
jonhoo: I think you could actually make this code a fair amount nicer by just sorting the batch first and then walking it in order. That way you'd know to not skip records whenever the key changes. You can take a look at how we do this [for grouped operators](https://github.com/mit-pdos/distributary/blob/67514028f32d94e78ce1ed2513fb963d94083281/dataflow/src/ops/grouped/mod.rs#L162L174). The implementation for [TopK](https://github.com/mit-pdos/distributary/blob/67514028f32d94e78ce1ed2513fb963d94083281/dataflow/src/ops/topk.rs#L131) is even closer to what you want to do; it also detects where groups start and stop, but the code is a little hairier as a result. Changing the code that way would also eliminate both `seen` and `cleaned` I think.

dataflow/src/ops/distinct.rs
```diff
@@ -0,0 +1,252 @@
...
+                    if self.group_by.iter().any(|col| col == &i) {
+                        Some(v)
+                    } else {
+                        None
+                    }
+                })
+                .cloned()
+                .collect::<Vec<_>>();
+            let positive = rec.is_positive();
+            if positive {
```
jonhoo: I'd probably put this `if` deeper inside so that we avoid duplicating the `db.lookup` code that is the same across both conditions.

dataflow/src/ops/distinct.rs
```diff
@@ -0,0 +1,252 @@
...
+                match db.lookup(group_by, &KeyType::from(&group[..])) {
+                    LookupResult::Some(rr) => {
+                        //println!("record {:?}", rr);
+                        if rr.len() == 0 {
+                            output.push(rec.clone());
+                        }
+                    },
+                    LookupResult::Missing => {
```
jonhoo: I would put `unimplemented!("Distinct does not yet support partial")` here so we don't miss (heh) this case should it happen due to a bug.

dataflow/src/ops/mod.rs
```diff
@@ -81,6 +84,8 @@ macro_rules! impl_ingredient_fn_mut {
             NodeOperator::TopK(ref mut i) => i.$fn($($arg),*),
             NodeOperator::Trigger(ref mut i) => i.$fn($($arg),*),
             NodeOperator::Rewrite(ref mut i) => i.$fn($($arg),*),
+            NodeOperator::Distinct(ref mut i) => i.$fn($($arg),*),
+
```
jonhoo: Spurious whitespace.

examples/basic-recipe.rs
```diff
@@ -30,8 +30,9 @@ fn main() {
 
     // set up Soup via recipe
     let mut builder = ControllerBuilder::default();
-    builder.log_with(distributary::logger_pls());
+    //builder.log_with(distributary::logger_pls());
     builder.set_worker_threads(2);
+    builder.disable_partial();
```
jonhoo: We should probably not change the provided example code this way :) Maybe make a test that executes this, and leave the example code the way it was?

mir/src/node.rs
```diff
@@ -871,7 +893,7 @@ impl Debug for MirNodeType {
                     .join(" ⋃ ");
 
                 write!(f, "{}", cols)
-            }
+            },
```
jonhoo: Spurious change.

src/controller/mir_to_flow.rs
```diff
@@ -739,6 +752,35 @@ pub(crate) fn make_project_node(
...
+    columns: &[Column],
+    group_by: &Vec<Column>,
+    mig: &mut Migration,
+) -> FlowNode {
+    let parent_na = parent.borrow().flow_node_addr().unwrap();
+    let column_names = column_names(columns);
+
+    let group_by_indx = if group_by.is_empty() {
+        // no query parameters, so we index on the first column
+        vec![0 as usize]
```
jonhoo: Shouldn't the default be to be distinct by *all* columns?

src/controller/sql/mir/grouped.rs
```diff
@@ -204,16 +204,16 @@ pub fn make_grouped(
...
                     group_cols.iter().collect(),
                     parent_node,
                 );
 
-                *prev_node = Some(n.clone());
-                func_nodes.push(n);
-                node_count += 1;
+                *prev_node = Some(nodes.last().unwrap().clone());
+                node_count += nodes.len();
+                func_nodes.extend(nodes);
```
jonhoo: This change seems unrelated to distinct?

jonhoo:
Nice! Congratulations on your first Soup operator :D I've left some comments inline, but overall looks pretty good. I'll leave the MIR changes for @ms705 to look at.

src/controller/mir_to_flow.rs
ms705: Agree, all columns should be the default here.

src/controller/sql/mir/mod.rs
```diff
@@ -805,19 +807,45 @@ impl SqlToMirConverter {
...
+                        name: a.clone(),
+                        alias: None,
+                        table: func_col.table.clone(),
+                        function: func_col.function.clone(),
+                    },
+                };
+                let mut dist_col = Vec::new();
+                dist_col.push(over);
+                dist_col.extend(group_cols.clone());
+                //println!("dist_cols {:?}", dist_col);
```
ms705: This can be dropped now, I think.

src/controller/sql/mir/mod.rs
```diff
@@ -805,19 +807,45 @@ impl SqlToMirConverter {
...
 
-        let mknode = |over: &Column, t: GroupedNodeType| {
-            self.make_grouped_node(name, &func_col, (parent, &over), group_cols, t)
+        let mut out_nodes = Vec::new();
+
+        let mknode = |over: &Column, t: GroupedNodeType, d: bool| {
+            if d {
+                let new_name = name.clone().to_owned() + "distinct";
+
+                let computed_col = match func_col.alias {
```
ms705: I think `computed_col` is not needed here, or is it?

src/controller/sql/mir/mod.rs
```diff
@@ -1581,6 +1629,22 @@ impl SqlToMirConverter {
                     new_node_count += 1;
                 }
 
+                info!(self.log, "distinct defined as {}", st.distinct);
```
ms705: I don't think we need this at `INFO` log level; drop or reduce?

src/controller/sql/mir/mod.rs
```diff
@@ -1581,6 +1629,22 @@ impl SqlToMirConverter {
...
+                if st.distinct {
+                    info!(self.log, "Adding distinct node to the query, distinct is set");
```
ms705: As above.

ms705:
Great work! :+1: 

MIR changes look good to me, modulo minor comments.

It'd be good to get this merged, since I plan to rewrite some of the MIR code that this PR modifies. @jsegaran are you likely to have time to finish it off? If not, @jonhoo or myself can take on rebasing, making the changes and merging it.

ms705:
Thanks for rebasing! I think this is now good to merge -- @jonhoo?

ms705:
Merge -- good work, @jsegaran!

PR #81, opened 2018-05-30
Worker recovery

ms705:
Initial implementation of worker recovery through recipe migrations.

PR for easy review of changes.


src/controller/domain_handle.rs
```diff
@@ -34,6 +34,7 @@ pub struct DomainHandle {
 
     cr_poll: PollingLoop<ControlReplyPacket>,
     shards: Vec<DomainShardHandle>,
+    nodes: Vec<NodeIndex>,
```
fintelia: It isn't clear whether we really want to track nodes here. I'm afraid it may be too easy for it to go out of sync with the controller's state.

src/controller/domain_handle.rs
```diff
@@ -186,6 +189,19 @@ impl DomainHandle {
         self.shards[shard].worker.clone()
     }
 
+    pub fn assigned_to_worker(&self, worker: &WorkerIdentifier) -> bool {
```
fintelia: This isn't the cleanest API, though I'm not sure of what might be better

src/controller/domain_handle.rs
```diff
@@ -209,6 +225,22 @@ impl DomainHandle {
         Ok(())
     }
 
+    pub(super) fn send_to_healthy(
```
fintelia: We should be more principled about sending to possibly failed nodes. At very least we should probably also use this function so we never send to nodes that we know are failed...

src/worker.rs
```diff
@@ -218,6 +225,8 @@ impl WorkerPool {
 
     pub fn wait(&mut self) {
         self.notify.clear();
+        self.wstate.exit.store(true, Ordering::SeqCst);
```
fintelia: In my experiments this didn't see to be fully working. I never got a chance to check why

ms705: What were the symptoms of it failing / how can I reproduce a failing case?

src/controller/domain_handle.rs
ms705: I agree, but it's tricky to get access to the controller state from within a `DomainHandle`.

Maybe we could make `assignments` available to the controller via an API, but that would still require the controller to be able to map the domain to its nodes. Is such a mapping available? (@jonhoo)

src/controller/domain_handle.rs
ms705: Agreed. I've marked the old `send` it as deprecated in 9796065, but got no compile warnings -- so I suspect it's already unused? (though that should also have produced warnings)

ms705: Never mind, I was being stupid (ae832ca). It's called in lots of places.

src/controller/domain_handle.rs
ms705: I think it's fine, at least for now. In the future, we might want to obtain the shard ID as a return value to selectively recovery that shard only.

src/controller/domain_handle.rs
ms705: I implemented a possible fix in 926f0f28bb6d77d4c828966257aa3c1163350343, keeping state only on the controller side. This has the effect of making recovery O(|D| * |V|), where D is the set of domains, and V is the set of data-flow vertices, however. This could be solved by keeping separate vertex->domain assignment state in `ControllerInner`.

ms705:
I think this is good to merge now, modulo some testing of the latest changes.

@fintelia @jonhoo

ms705:
I retract my above assessment -- this is still buggy.

Specifically, although recovery works fine, the dead nodes are never actually removed from the graph. I'm working on a fix.

src/worker.rs
fintelia: The worker would not actually stop, and that function would wait forever.

PR #82, opened 2018-09-13
Functional Piazza example with security universes. 

alanamarzoev:


mir/src/node.rs
```diff
@@ -227,16 +228,86 @@ impl MirNode {
...
-                        c, self.name, self.columns
-                    );
-                }
-                Some(id) => id,
+                    // See if table mapping was passed in
+                    match table_mapping {
+                        // if mapping was passed in, then see if c has an associated table, and check
+                        // the mapping for a key based on this
+                        Some(map) => {
+                            match c.table.clone() {
```
ms705:
```suggestion
                            match *c.table {
```
... should avoid the `clone`.

mir/src/node.rs
```diff
@@ -227,16 +228,86 @@ impl MirNode {
...
+                                        Some(ref mut t_name) => {
+
+                                            let mut ind = 0;
+                                            for self_column in &self.columns {
+                                                if self_column.name == c.name && self_column.table == Some(t_name.clone()) {
+                                                    return ind;
+                                                }
+
+                                                ind = ind + 1;
+                                            }
```
ms705: You can replace this loop by 
```rust
let mut ac = c.clone();
ac.table = Some(t_name.to_owned());
let ind = self.columns.position(|cc| cc == ac).expect("tried to ...");
```
or something similar, if I'm not mistaken, as the loop only computes the index for a matching column.

mir/src/node.rs
```diff
@@ -227,16 +228,86 @@ impl MirNode {
...
+                                            panic!("tried to look up non-existent column {:?} on node \"{}\" (columns: {:?})",
+                                                   c, self.name, self.columns);
+                                        },
+                                        None => {
+                                            let mut cn = c.clone().name;
+                                            match map.get(&cn) {
+                                                Some(ref mut res) => {
+                                                    let mut ind = 0;
+                                                    for self_column in &self.columns {
```
ms705: See above re loop.

mir/src/node.rs
```diff
@@ -227,16 +228,86 @@ impl MirNode {
...
+                                                }
+
+                                                ind = ind + 1;
+                                            }
+                                            panic!("tried to look up non-existent column {:?} on node \"{}\" (columns: {:?})",
+                                                   c, self.name, self.columns);
+                                        },
+                                        None => {
+                                            let mut cn = c.clone().name;
+                                            match map.get(&cn) {
```
ms705: I'm curious to understand the design better here -- looks like there are *two* lookups into the mapping table, one for `column:table` and one just for `column`. Why do we need both?

mir/src/node.rs
```diff
@@ -227,16 +228,86 @@ impl MirNode {
...
+                                            }
+                                        }
+                                    }
+                                },
+                                None => {
+                                    let keyed = c.clone().name;
+                                    match map.get(&keyed) {
+                                        Some(ref mut t_n) => {
+                                            let mut ind = 0;
+                                            for self_column in &self.columns {
```
ms705: See above re loop

mir/src/node.rs
```diff
@@ -227,16 +228,86 @@ impl MirNode {
...
+                                                   c, self.name, self.columns);
+                                        },
+                                        None => {
+                                            panic!("tried to look up non-existent column {:?} on node \"{}\" (columns: {:?})",
+                                                   c, self.name, self.columns);
+                                        }
+                                    }
+                                }
+                            }
+                        },
```
ms705: The above has a fair amount of repetition; I wonder if we could pull out the code that computes the column index and panics if the column doesn't exist, as it's identical except for the column name/table that we're looking for.
I'm thinking:
```rust
let needed_col = match *table_mapping { /* complex multi-case statement */ };
let ind = self.columns.position(|cc| cc = needed_col).expect("tried to ...");
```

src/controller/sql/mir/mod.rs
```diff
@@ -1332,489 +1435,488 @@ impl SqlToMirConverter {
...
-        qg: &QueryGraph,
-        has_leaf: bool,
-        universe: UniverseId,
-    ) -> Vec<MirNodeRef> {
-        use crate::controller::sql::mir::grouped::make_grouped;
-        use crate::controller::sql::mir::grouped::make_predicates_above_grouped;
-        use crate::controller::sql::mir::join::make_joins;
-        use std::collections::HashMap;
+    /// Returns list of nodes added
+fn make_nodes_for_selection(
```
ms705: Was this accidentally unindented, or is it a deliberate change?

src/controller/sql/mod.rs
```diff
@@ -526,30 +526,42 @@ impl SqlIncorporator {
...
             is_leaf,
             universe.clone(),
         );
 
-        trace!(self.log, "Unoptimized MIR:\n{}", mir.to_graphviz().unwrap());
+        trace!(self.log, "Unoptimized MIR:\n{}", og_mir.to_graphviz().unwrap());
 
         // run MIR-level optimizations
-        mir = mir.optimize();
+        let mut mir = og_mir.optimize(table_mapping.clone(), sec);
```
ms705: Let's use a ref to avoid this clone

src/controller/sql/mod.rs
```diff
@@ -526,30 +526,42 @@ impl SqlIncorporator {
...
 
         trace!(self.log, "Optimized MIR:\n{}", mir.to_graphviz().unwrap());
 
+        if sec {
+            match table_mapping.clone() {
```
ms705: No clone needed here, I think?  `table_mapping` isn't used again, so we should be able to take ownership

src/controller/sql/mod.rs
```diff
@@ -703,13 +716,26 @@ impl SqlIncorporator {
 
         let mut post_reuse_opt_mir = reused_mir.optimize_post_reuse();
 
+        // traverse universe subgraph and update table names for
+        // internal consistency using the table mapping as guidance
+        if sec {
+            match table_mapping.clone() {
```
ms705: Avoid clone using `Some(ref x)`

mir/src/node.rs
```diff
@@ -205,7 +206,7 @@ impl MirNode {
         self.columns.as_slice()
     }
 
-    pub fn column_id_for_column(&self, c: &Column) -> usize {
+    pub fn column_id_for_column(&self, c: &Column, table_mapping: Option<HashMap<String,String>>) -> usize {
```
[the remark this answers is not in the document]
ms705: Done

mir/src/node.rs
```diff
@@ -227,16 +228,86 @@ impl MirNode {
                     .1
                     .expect("must have an absolute column ID on base"),
             },
-            MirNodeType::Reuse { ref node } => node.borrow().column_id_for_column(c),
+            MirNodeType::Reuse { ref node } => node.borrow().column_id_for_column(c, table_mapping.clone()),
```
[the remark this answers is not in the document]
ms705: Done

mir/src/node.rs
ms705: Done, but needed `Some(ref map)` instead

ms705:
I've fixed the easy comments related to ownership/clone, indentation changes, etc., merged with the latest `master`, and fixed conflicts (not many, fortunately).

The remaining changes relate to indexing the `table_mappings` map by `(String, String)` and some loops (see comments above). Once we've fixed those, I think we should merge this -- it's been a long time coming!

PR #113, opened 2018-11-02
Static reader node replication

ygina:
Place replicas in separate domains, including when there is only 1 replica. Place these domains round robin on separate workers. (`migrate/mod.rs`)
Support for removing all reader replicas along with a view. (`controller/inner.rs`)
Obtain views to replicas round robin. CLI options to set replicas and reader threads.

**Benchmarks:** 16 client threads and cores, 4 Noria servers with 1 core per server, 1 write every 100 on the vote benchmark. As number of replicas increases, Noria can handle a greater number of ops/s before falling behind.
**Tests:** Same as before, with updated node counts to account for additional replicas. One new test to check that view actually returns different replicas, and that writes are reflected in these replicas.

### DIFF
--- a/noria-server/src/controller/inner.rs
+++ b/noria-server/src/controller/inner.rs
@@ -1145,35 +1145,19 @@ impl ControllerInner {
         // We're looking for a single egress node that connects the query node to readers in
         // other domains.
         let mut nodes = vec![];
-        let egress_node = {
-            let num_children = self
-                .ingredients
+        let num_children = self
+            .ingredients
+            .neighbors_directed(leaf, petgraph::EdgeDirection::Outgoing)
+            .count();
+        if num_children == 1 {
+            let child = self.ingredients
                 .neighbors_directed(leaf, petgraph::EdgeDirection::Outgoing)
-                .count();
-            if num_children == 0 {
-                None
-            } else if num_children == 1 {
-                let child = self.ingredients
-                    .neighbors_directed(leaf, petgraph::EdgeDirection::Outgoing)
-                    .next()
-                    .unwrap();
-                assert!(self.ingredients[child].is_egress());
-                Some(child)
-            } else {
-                // should not happen, since we remove nodes in reverse topological order
-                crit!(
-                    self.log,
-                    "not removing node {} yet, as it still has non-reader-related children",
-                    leaf.index();
-                    "num_children" => num_children,
-                );
-                unreachable!();
-            }
-        };
+                .next()
+                .unwrap();
+            assert!(self.ingredients[child].is_egress());
 
-        // Remove the egress node and its children
-        if let Some(egress_node) = egress_node {
-            let mut bfs = Bfs::new(&self.ingredients, egress_node);
+            // Remove the egress node and its children
+            let mut bfs = Bfs::new(&self.ingredients, child);
             while let Some(child) = bfs.next(&self.ingredients) {
                 if self.ingredients
                     .neighbors_directed(child, petgraph::EdgeDirection::Outgoing)
@@ -1185,19 +1169,18 @@ impl ControllerInner {
             }
             debug!(
                 self.log, "Removing egress node and its children";
-                "node" => egress_node.index(),
+                "node" => child.index(),
                 "leaf" => leaf.index(),
             );
-        }
-
-        // The nodes we remove first do not have children any more
-        for node in &nodes {
-            assert_eq!(
-                self.ingredients
-                    .neighbors_directed(*node, petgraph::EdgeDirection::Outgoing)
-                    .count(),
-                0
+        } else if num_children > 1 {
+            // should not happen, since we remove nodes in reverse topological order
+            crit!(
+                self.log,
+                "not removing node {} yet, as it still has non-reader-related children",
+                leaf.index();
+                "num_children" => num_children,
             );
+            unreachable!();
         }
 
         while let Some(node) = nodes.pop() {


noria-server/src/controller/inner.rs
```diff
@@ -533,17 +533,10 @@ impl ControllerInner {
                 persistence_parameters: self.persistence.clone(),
             };
 
-            let (identifier, w) = loop {
-                if let Some((i, w)) = wi.next() {
-                    if w.healthy {
-                        break (*i, w);
```
jonhoo: I wonder why this old code `break`s here? And why the new code doesn't have to?

noria-server/src/controller/inner.rs
```diff
@@ -533,17 +533,10 @@ impl ControllerInner {
...
-                    }
-                } else {
-                    wi = self.workers.iter_mut();
-                }
-            };
-
             // send domain to worker
+            let identifier = identifiers.get(i)
+                .expect("number of identifiers should match number of shards");
```
jonhoo: Doesn't this need to take into account whether workers are healthy (`w.healthy`)? What if there are more shards than workers?

noria-server/src/controller/inner.rs
```diff
@@ -721,30 +714,12 @@ impl ControllerInner {
...
-        }
-
-        reader
-    }
-
     /// Obtain a `ViewBuilder` that can be sent to a client and then used to query a given
-    /// (already maintained) reader node called `name`.
-    pub fn view_builder(&self, name: &str) -> Option<ViewBuilder> {
+    /// (already maintained) reader node. If the view has replicas, a `ViewBuilder` is returned
+    /// for each view in round robin order.
```
jonhoo: for each _replica_? I wonder if there's a way we can clear up the language here and elsewhere... In particular, it seems like sometimes refer to "replicas" as all the _other_ `Reader`s except the "first one", but we also store all the `Reader`s in a single `Vec` called `replicas`. Which is it? Maybe the way way to talk about this is that a "view" has one or more "instances", and the instances are all replicas of one another. "Instance" then maps directly onto `Reader`. Sadly we use the word "instance" in all sorts of settings, so we may need something less common. "Manifestation" perhaps? It could be that the word we actually want to use here is "materialization" (i.e., a given view may have multiple (identical) materializations). @ms705 ?

noria-server/src/controller/inner.rs
```diff
@@ -756,16 +731,29 @@ impl ControllerInner {
             }
         };
 
-        self.find_view_for(node).map(|r| {
-            let domain = self.ingredients[r].domain();
-            let columns = self.ingredients[r].fields().to_vec();
+        // reader should be a child of the given node. however, due to sharding and replication,
+        // it may not be an *immediate* child. furthermore, once we go beyond depth 1, we may
+        // accidentally hit an *unrelated* reader node. to account for this, nodes cache their
+        // readers so we can easily query data.
```
jonhoo: I think this comment should now go away, since we no longer have to look past ingress/egress nodes here.

noria-server/src/controller/inner.rs
```diff
@@ -756,16 +731,29 @@ impl ControllerInner {
...
+        self.ingredients[node].next_replica().map(|ni| {
+            let rname = self.ingredients[ni].name();
```
jonhoo: You could just do `let name = ...to_string();` directly here; that way you can use the short-form of just `name,` when you make the `ViewBuilder` below.

noria-server/src/controller/inner.rs
```diff
@@ -1109,7 +1097,7 @@ impl ControllerInner {
         graphviz(&self.ingredients, detailed, &self.materializations)
     }
 
-    fn remove_leaf(&mut self, mut leaf: NodeIndex) -> Result<(), String> {
+    fn remove_query_node(&mut self, leaf: NodeIndex) -> Result<(), String> {
```
jonhoo: @fintelia can you take a look at the change to this method? I'm not familiar with what it does and what invariants caller may assume.

noria-server/src/controller/migrate/assignment.rs
```diff
@@ -109,6 +110,11 @@ pub fn assign(
                 };
             }
 
+            // replicas are always in their own domain to distribute the load.
+            if let Some(_) = n.replica_index() {
```
jonhoo: This is one of the places where language clarification would help. Is the first `Reader` also a replica? If not, then surely it shouldn't have a `replica_index`? But it does, because currently the first `Reader` is also placed in its own domain?

noria-server/src/controller/migrate/mod.rs
```diff
@@ -205,47 +205,204 @@ impl<'a> Migration<'a> {
...
-            let r = node::special::Reader::new(n);
-            let r = if let Some(name) = name {
-                self.mainline.ingredients[n].named_mirror(r, name)
-            } else {
-                self.mainline.ingredients[n].mirror(r)
-            };
-            let r = self.mainline.ingredients.add_node(r);
-            self.mainline.ingredients.add_edge(n, r, ());
-            self.readers.insert(n, r);
+            let mut readers = vec![];
```
jonhoo:
```suggestion
            let mut readers = Vec::with_capacity(num_replicas);
```

noria-server/src/controller/migrate/mod.rs
```diff
@@ -205,47 +205,204 @@ impl<'a> Migration<'a> {
...
+            for i in 0..num_replicas {
+                let replica_i = {
+                    if num_replicas == 0 {
+                        None
+                    } else {
+                        Some(i)
+                    }
```
jonhoo: Is this distinction really important? Do we need to distinguish between a view with one materialization and with a replication factor of 1 (which would also just have one materialization)?

noria-server/src/controller/migrate/mod.rs
```diff
@@ -205,47 +205,204 @@ impl<'a> Migration<'a> {
...
+                };
+
+                // Make a reader node
+                let r = node::special::Reader::new(n, replica_i);
+                let r = if let Some(name) = name.clone() {
+                    if i == 0 {
+                        self.mainline.ingredients[n].named_mirror(r, name)
+                    } else {
+                        self.mainline.ingredients[n].named_mirror(r, format!("{}_r{}", name, i))
```
jonhoo: TBH I'd just all name them this way. Maybe cut the `r`?

noria-server/src/controller/migrate/mod.rs
```diff
@@ -205,47 +205,204 @@ impl<'a> Migration<'a> {
...
     /// To query into the maintained state, use `ControllerInner::get_getter`.
     #[cfg(test)]
-    pub fn maintain_anonymous(&mut self, n: NodeIndex, key: &[usize]) -> NodeIndex {
-        self.ensure_reader_for(n, None);
-        let ri = self.readers[&n];
-
-        self.mainline.ingredients[ri]
-            .with_reader_mut(|r| r.set_key(key))
-            .unwrap();
+    pub fn maintain_anonymous(&mut self, n: NodeIndex, key: &[usize]) -> Vec<NodeIndex> {
```
jonhoo: I feel like we should just kill this method and mandate that people use `maintain`...

noria-server/src/controller/migrate/mod.rs
```diff
@@ -333,16 +492,11 @@ impl<'a> Migration<'a> {
             }
         }
         let swapped = swapped0;
+
+        // Assign local addresses to all new nodes, and initialize them.
         let mut sorted_new = new.iter().collect::<Vec<_>>();
```
jonhoo: You may want to add a `map` here so that this becomes a `Vec<NodeIndex>`. That way you don't have to have all the `&&&&&&&&` below :p

noria-server/src/controller/migrate/mod.rs
```diff
@@ -451,24 +542,93 @@ impl<'a> Migration<'a> {
...
-                // this is not a new domain
-                continue;
+        // Bucket reader replica nodes and the non-replica nodes separately.
+        let mut replica_map = HashMap::new();
+        let mut other_nodes = HashSet::new();
+        for &&ni in sorted_new
+                .iter()
+                .filter(|&&&ni| ni != mainline.source)
+                .filter(|&&&ni| !mainline.ingredients[ni].is_dropped())
+                .into_iter() {
```
jonhoo: You don't need the `into_iter` here.

noria-server/src/controller/migrate/mod.rs
jonhoo: I also wonder if we should just filter on `!is_dropped()` here, since afaict we add that filter every time we use `sorted_new` below.

noria-server/src/controller/migrate/mod.rs
```diff
@@ -451,24 +542,93 @@ impl<'a> Migration<'a> {
...
+            // Check if the node is a reader, and then insert it in the replica map. The value
+            // is a list of readers that are for the same node, and the key is the node the
+            // readers are for. Later, we will also add the associated ingress node to the replica
+            // map, since readers are all located on their own domains.
+            if let Some(for_node) = mainline.ingredients[ni]
+                    .with_reader(|r| Some(r.is_for()))
+                    .unwrap_or(None) {
```
jonhoo: I'm confused... Can't you just use `self.readers` here instead?

noria-server/src/controller/migrate/mod.rs
```diff
@@ -451,24 +542,93 @@ impl<'a> Migration<'a> {
...
+        // Obtain a vec of DomainIndexes in the order that we want to assign them to workers.
+        // For replicas, all replicas for the same reader should be consecutive. For non-replicas,
+        // it doesn't really matter unless we want the ordering to be deterministic.
+        let mut changed_domains_replicas = Vec::new();
+        for (_, readers) in &mut replica_map {
+            let mut domains = readers
+                .iter()
+                .filter(|&ni| !mainline.ingredients[*ni].is_dropped())
+                .map(|&ni| mainline.ingredients[ni].domain())
+                .collect::<Vec<_>>();
```
jonhoo: Why do you need to `collect` here if you're then immediately `append`ing to `changed_domains_replicas`? Couldn't you just give the iterator directly to `.extend` instead?

noria-server/src/controller/migrate/mod.rs
```diff
@@ -451,24 +542,93 @@ impl<'a> Migration<'a> {
...
 
-            let nodes = uninformed_domain_nodes.remove(&domain).unwrap();
-            let d = mainline.place_domain(
-                domain,
-                mainline.ingredients[nodes[0].0].sharded_by().shards(),
-                &log,
-                nodes,
+        // Obtain a vec of DomainIndexes in the order that we want to assign them to workers.
+        // For replicas, all replicas for the same reader should be consecutive. For non-replicas,
+        // it doesn't really matter unless we want the ordering to be deterministic.
```
jonhoo: This code below is pretty complicated, and it's not entirely clear why or how it does what it does. Could you try to simplify it?

noria/src/controller.rs
```diff
@@ -224,13 +225,10 @@ impl<A: Authority> ControllerHandle<A> {
                 if let Some(port) = self.local_port {
                     g = g.with_local_port(port);
                 }
-
```
jonhoo: Is there a particular reason for these whitespace changes?

noria/src/view.rs
```diff
@@ -182,6 +188,11 @@ impl View<SharedConnection> {
     allow(clippy::len_without_is_empty)
 )]
 impl<E> View<E> {
+    /// Get the name of the corresponding reader node.
+    pub fn name(&self) -> String {
```
jonhoo:
```suggestion
    pub fn name(&self) -> &str {
```

noria/src/view.rs
```diff
@@ -59,6 +59,7 @@ pub enum ReadReply {
 #[doc(hidden)]
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ViewBuilder {
+    pub name: String,
```
jonhoo: Why do views now need to carry their own name? I don't think it's a bad change, just curious.

noria-server/src/controller/migrate/mod.rs
ygina: 👍

ygina:
Going to simplify `commit()` and clarify the language surrounding replicas to address the remaining comments. It seems like sharding followed the convention where None meant 1 shard. For replication, i think it makes sense to think of the replication parameter as the "replication factor" rather than the number of replicas _in addition_ to the original reader --> so replication factor of 1 = 1 reader.

noria-server/dataflow/src/node/mod.rs
```diff
@@ -110,9 +110,9 @@ impl Node {
         self.sharded_by = s;
     }
 
-    /// Add a reader replica to this node.
-    pub fn add_replica(&mut self, ni: NodeIndex) {
-        self.replicas.push(ni)
+    /// The node being added is a reader for this node.
```
jonhoo: What is "the node being added"? Do you mean `ni`?

noria-server/dataflow/src/node/mod.rs
```diff
@@ -342,32 +342,28 @@ impl Node {
...
-    /// Returns replicas in round robin order each time the method is called.
-    pub fn next_replica(&mut self) -> Option<NodeIndex> {
-        if self.num_replicas() > 0 {
-            self.replica_index += 1;
-            self.replica_index %= self.num_replicas();
-            Some(*self.replicas.get(self.replica_index).unwrap())
+    /// Returns reader replicas in round robin order each time the method is called.
+    pub fn next_reader(&mut self) -> Option<NodeIndex> {
+        if self.num_readers() > 0 {
+            self.next_reader += 1;
```
jonhoo: Interesting... I just realized that this means we'll give out the *first* reader *last*. Maybe that's what we want since the first reader is on the same machine as the rest of the write path, which is likely to be more loaded? I don't know.

noria-server/dataflow/src/node/special/reader.rs
```diff
@@ -35,7 +35,7 @@ pub struct Reader {
     streamers: Vec<channel::StreamSender<Vec<StreamUpdate>>>,
 
     for_node: NodeIndex,
-    replica_index: Option<usize>,
+    reader_index: usize,
```
jonhoo: What does this field mean? Is it the index of this `Reader` in the list of `Reader`s for its ancestor?

noria-server/src/controller/builder.rs
```diff
@@ -57,7 +57,7 @@ impl ControllerBuilder {
         self.config.sharding = shards;
     }
 
-    /// Set number of reader replicas
+    /// Set reader replication factor
     pub fn set_replicas(&mut self, replicas: usize) {
```
jonhoo: Maybe rename this to `set_replication_factor`?

noria-server/src/controller/inner.rs
```diff
@@ -720,10 +720,10 @@ impl ControllerInner {
     }
 
     /// Obtain a `ViewBuilder` that can be sent to a client and then used to query a given
-    /// (already maintained) reader node. If the view has replicas, a `ViewBuilder` is returned
-    /// for each view in round robin order.
+    /// (already maintained) reader node. If there are multiple readers for any given query,
+    /// a `ViewBuilder` is returned for each reader in round robin order.
     ///
-    /// The name of the reader node is `name`, or if it is a replica, `name_r[REPLICA_NUMBER]`.
+    /// `name` is the name of the query.
```
jonhoo: Isn't name the name of the `view` that the `Reader` should be constructed for?

noria-server/src/controller/inner.rs
```diff
@@ -732,11 +732,18 @@ impl ControllerInner {
...
                 // if the recipe doesn't know about this query, traverse the graph.
                 // we need this do deal with manually constructed graphs (e.g., in tests).
-                *self.outputs().get(name)?
+                if let Some(ni) = self.outputs().get(name) {
+                    *ni
+                } else {
+                    // depending on how the graph was constructed, the outputs may be suffixed
+                    // with the reader index.
+                    let reader_name = format!("{}_0", name);
+                    *self.outputs().get(&reader_name)?
```
jonhoo: Doesn't this mean that tests will only ever use the first replica? Why is it correct to just always use `_0` here?

noria-server/src/controller/inner.rs
```diff
@@ -1117,10 +1124,11 @@ impl ControllerInner {
...
         {
-            // This query leaf node has children -- typically, these are readers, but they can
-            // also be egress nodes or other, dependent queries. If the reader has replicas, we
-            // are looking for a single egress node that connects the leaf to replicas in other
-            // domains.
+            // This query leaf node has children -- egress nodes or other, dependent queries.
+            // Typically, we are looking for a single egress node that connects the leaf to
+            // readers in other domains.
+            // TODO: if each reader (including the 1st) definitely gets its own domain, this
+            // code can be a lot simpler.
```
jonhoo: How much is "a lot"? I think we should favor simplicity then.

noria-server/src/controller/migrate/mod.rs
```diff
@@ -209,35 +209,23 @@ impl<'a> Migration<'a> {
...
-                        None
-                    } else {
-                        Some(i)
-                    }
-                };
-
                 // Make a reader node
-                let r = node::special::Reader::new(n, replica_i);
+                let r = node::special::Reader::new(n, i);
                 let r = if let Some(name) = name.clone() {
```
jonhoo: Why does this need to `clone`, given that it's just passed to `format!` below?

jonhoo: Ah, it did in the old code, since it used `name` directly, but I don't think it needs to any longer.

noria-server/src/controller/inner.rs
jonhoo: Ping @fintelia

noria-server/src/controller/migrate/mod.rs
jonhoo: @fintelia @ms705 thoughts on this?

noria-server/src/controller/migrate/mod.rs
jonhoo: Was this actually resolved? I think you are still missing a `map(|&ni| ni)`?

noria-server/src/controller/migrate/mod.rs
jonhoo: That helps a little, though I'm still not sure why we need to create *two* `changed_domains_*` and then call `place_round_robin` twice, as opposed to just having `place_round_robin` know about what things are readers and what things are not, and assign appropriately in a single pass?

noria/src/view.rs
jonhoo: I think it might make sense for `ViewBuilder` to know which replica (i.e., `usize`) it is interacting with, especially when we later add fail-over. Maybe just add that field?

noria-server/src/controller/migrate/mod.rs
ygina: 👍

noria-server/src/controller/builder.rs
ygina: 👍

noria-server/src/controller/inner.rs
ygina: `outputs()` returns a map from reader name to "the the node address that is being materialized not that of the reader node itself". So for something like `it_works_with_simple_arithmetic()` all of `CarPrice_0`, `CarPrice_1`, `CarPrice_2` map to the same node. From there we choose a replica from the list of replicas cached in the node. In a test like `add_columns()` I guess the graph is constructed manually(?) and the readers have a name like `a` instead of `a_0`.

noria-server/src/controller/inner.rs
ygina: What's the difference between queries and external views?

noria-server/dataflow/src/node/special/reader.rs
ygina: Yes, added a comment.

noria-server/dataflow/src/node/mod.rs
ygina: Either works! Certainly wasn't intentional.

noria-server/dataflow/src/node/mod.rs
ygina: Yes, modified the comment.

noria-server/src/controller/migrate/mod.rs
ygina: 👍

noria-server/src/controller/migrate/mod.rs
ygina: I could assign domains to replicas more like how it's done for shards, but I'm not actually sure how domain assignment works for shards -- seems like it's ok for shards to share domains with other nodes?

Since we're placing round robin, we have to specifically list readers for the same node in consecutive order in `changed_domain_readers`, to ensure their respective domains are placed on different workers. To only call `place_round_robin` once, we _could_ combine `changed_domain_*` so long as the readers are still in the right order, or move the reader-distinguishing logic into `place_round_robin`. Did you have something specific in mind?

noria-server/src/controller/inner.rs
jonhoo: Ohhhhhh, I see what you mean. You're not actually looking for the `Reader` here, but their ancestor. Makes sense.

noria-server/src/controller/inner.rs
jonhoo: Everything SQL statement the system is given is a query. We only add `Reader`s for those queries that the application explicitly chooses to make "public" using `CREATE VIEW` (which in turn calls `maintain`)

noria-server/dataflow/src/node/mod.rs
jonhoo: Maybe add it to the documentation for `next_reader`: "with the primary reader being returned last"?

noria/src/view.rs
ygina: 👍

noria-server/src/controller/migrate/mod.rs
jonhoo: For shards, the logic is pretty straightforward, since any domain that is sharded simply gets "placed" as many times as there are shards ([here](https://github.com/mit-pdos/noria/pull/113/files#diff-c2efd5de7d5bd366a1b53929183e8c12R374)). I guess for replicas, this is harder since you don't have the same "one thing is many things" to rely on in `place`. I think it isn't so much that I think the approach you've taken is wrong or weird, it's just that I didn't fully understand it when I first read the code. It could be that a comment at the top about why we need to invoke `place_round_robin` twice would be enough.

noria-server/src/controller/inner.rs
ygina: 👍

noria-server/src/controller/inner.rs
ms705: Side-note: there are two other ways to get "public" queries with a reader -- using `VIEW <name>: SELECT ...` and `QUERY <name>: SELECT ...` in the recipe. Both end up calling the same code paths (`maintain`, etc.), though.

jonhoo:
Given 232453f9bbd94383af312d432f5a41766f10f725, I'd particularly like @fintelia to take a look at this.

noria-server/src/controller/inner.rs
fintelia: I don't think these changes are right. This code will panic if the node to be removed does not have exactly one child, but I think it should actually be asserting that the indicated node has no non-reader children. I'm also not sure I understand why the function name was changed.

ygina: I changed the structure of the graph such that views are connected to readers through egress/ingress nodes, so these nodes shouldn't be directly connected to readers anymore.

I might have misunderstood the function purpose -- I changed the name since I thought removing a "leaf" implied the leaf had no children, which is not the case. Rather the node passed in is the query node with an egress node as a child (used to be a reader as a child). I'm not quite sure I understand the old comment about how the query leaf node may have "egress nodes or other, dependent queries" as children, but the code panicked if `has_non_reader_children` was true.

noria-server/src/controller/migrate/mod.rs
fintelia: Seems reasonable.

noria-server/src/controller/inner.rs
ygina: I see, so the function _should_ only be called on leaf nodes of the MIR graph, and should panic otherwise. Sometimes those leaf nodes are query nodes and sometimes they are just other nodes being removed in topological order (no children). So I should modify the code to account for the second case. Does that sound correct?

fintelia: Yeah, that sounds right. You might also want to check with @ms705 to make sure I'm recalling the MIR graph part correctly.

ms705:
Where did we end up with this? I vaguely remember discussing the outstanding issues with @fintelia before the holidays.

fintelia:
Last I recall query removal was being corrected because it didn't properly handle deletion of non-reader nodes. I'm not sure what the current status is

ygina:
@fintelia I don't have any more changes to make. Is there a case you're concerned about in particular? It looks like the old code removes non-reader nodes in reverse topological order before it reaches my code for removing leaf/reader nodes, so it shouldn't be affected.

noria-server/src/controller/inner.rs
```diff
@@ -1146,60 +1142,64 @@ impl ControllerInner {
...
-        // `node` now does not have any children any more
-        assert_eq!(
-            self.ingredients
-                .neighbors_directed(leaf, petgraph::EdgeDirection::Outgoing)
-                .count(),
-            0
-        );
+        // The nodes we remove first do not have children any more
+        for node in &nodes {
+            assert_eq!(
```
fintelia: Couldn't this check have been done when node is added to the collection?

noria-server/src/controller/inner.rs
```diff
@@ -1146,60 +1142,64 @@ impl ControllerInner {
...
+                .count();
+            if num_children == 0 {
+                None
+            } else if num_children == 1 {
+                let child = self.ingredients
+                    .neighbors_directed(leaf, petgraph::EdgeDirection::Outgoing)
+                    .next()
+                    .unwrap();
+                assert!(self.ingredients[child].is_egress());
+                Some(child)
```
fintelia: This is the only branch that returns `Some()`. Why not just place the if statement from below right here instead?

Uniminin (approved):
looks good!

PR #117, opened 2019-02-21
join chain generation

jmbredenberg:
Fixes join chain code (as I understand it, it was previously not noticing if two desired tables were already in the same chain) and adds some tests to ensure correct behavior.


noria-server/src/controller/sql/mir/join.rs
```diff
@@ -83,7 +92,7 @@ fn pick_join_chains(
     dst: &String,
```
ms705: Not directly related to the changes in this PR, but this can be `&str`, and changing it will avoid the `&"a".to_string()` in your test code.

noria-server/src/controller/sql/mir/join.rs
```diff
@@ -100,5 +113,184 @@ fn pick_join_chains(
...
+                assert!(new_chain.has_table(&"D".to_string()));
+                join_chains.push(new_chain);
+            },
+            None => {
+                assert!(false);
+            },
+        };
+
+        // now query a different combination of them
+        let (left_chain, second_chain) = pick_join_chains(&"A".to_string(), &"C".to_string(), &mut join_chains, &node_for_rel);
```
ms705: I'm trying to think through whether this test is correct: if I understand correctly, it says "I'm trying to join tables A and C, do we have any join chains for that?", and the result is the chain with A, B, C, and D. But is an ABCD chain really usable for an AC join?`What if the chain (as in this test) first joins A and B and then joins the result with C? The rows in the ABC output aren't necessarily the same as those in an AC join (in particular, there may be more or fewer rows depending on what rows B has).

Now, if the join were ACB, that's different: the chain works because we can just reuse the first two nodes. But `JoinChain` has no notion of table ordering, so we can't tell which one it is.

noria-server/src/controller/sql/mir/join.rs
```diff
@@ -46,24 +48,31 @@ pub fn make_joins(
...
+                join_chains.push(new_chain);
 
-        node_count += 1;
+                node_count += 1;
 
-        join_nodes.push(jn);
+                join_nodes.push(jn);
+            },
+            None => {
+                join_chains.push(left_chain);
```
ms705: Putting back `left_chain` in this case is definitely more correct than what we had before, but I wonder if we need to be careful about where to put it in the vector.

`pick_join_chains` searches the vector of chains in order, so we'll pull out the first chain that has all the tables we're looking for. That may be a longer chain than we're looking for (e.g., we might find ABC when a later entry in the vector has AB). I _think_ this is fine because later code will scan along the longer chain and find the AB join without caring about the extra C, but can you double-check that this is the case?

ms705:
Closing this as we've abandoned the change for now.

PR #120, opened 2019-04-01
Type: fix link in readme

alexsnaps:
This is straight forward, but makes getting up to speed less of a burden :)


jonhoo:
Perfect, thank you!

PR #124, opened 2019-04-28
Move back to upstream rust-rocksdb

ekmartin:
The only significant change here is that `ColumnFamily` now [has a lifetime parameter](https://github.com/rust-rocksdb/rust-rocksdb/commit/248b40b4659c4f300b46a0948b3fc175d321c397) which makes it harder to store as a part of `PersistentState`. I initially threaded through the lifetime parameter up to `PersistentState` but couldn't get past this error:

`error[E0277]: *mut librocksdb_sys::rocksdb_column_family_handle_t cannot be shared between threads safely --> noria-server/dataflow/src/state/persistent_state.rs:70:10`

`ColumnFamily` is marked as `Send` but I wonder if the way to solve this might be to mark the [inner pointer](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/src/lib.rs#L288) as [`NonNull`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html) in rust-rocksdb? Not sure.

To work around it I changed the column family callsites to instead use `db.cf_handle(name)`, which takes a lock (added in https://github.com/rust-rocksdb/rust-rocksdb/pull/197) and reads from a `BTreeMap`.

### DIFF
--- a/noria-server/dataflow/Cargo.toml
+++ b/noria-server/dataflow/Cargo.toml
@@ -35,6 +35,7 @@ common = { path = "../common" }
 noria = { path = "../../noria" }
 
 [dependencies.rocksdb]
-git = "https://github.com/ekmartin/rust-rocksdb.git"
+git = "https://github.com/rust-rocksdb/rust-rocksdb.git"
+default-features = false
 features = ["lz4"]
-branch = "custom"
+rev = "184440b2095979baeb747ccf300022198e9606ed"
--- a/noria-server/dataflow/Cargo.toml
+++ b/noria-server/dataflow/Cargo.toml
@@ -29,13 +29,8 @@ backtrace = { version = "0.3.2", features = ["serialize-serde"] }
 petgraph = { version = "0.4.11", features = ["serde-1"] }
 serde = { version = "1.0.8", features = ["rc"] }
 timekeeper = {version = "0.3.0", default-features = false }
+rocksdb = {version = "0.12.2", default-features = false, features = ["lz4"] }
 
 # local deps
 common = { path = "../common" }
 noria = { path = "../../noria" }
-
-[dependencies.rocksdb]
-git = "https://github.com/rust-rocksdb/rust-rocksdb.git"
-default-features = false
-features = ["lz4"]
-rev = "184440b2095979baeb747ccf300022198e9606ed"
--- a/noria-server/dataflow/src/state/persistent_state.rs
+++ b/noria-server/dataflow/src/state/persistent_state.rs
@@ -1,6 +1,6 @@
 use bincode;
 use itertools::Itertools;
-use rocksdb::{self, ColumnFamily, SliceTransform, SliceTransformFns, WriteBatch};
+use rocksdb::{self, PlainTableFactoryOptions, SliceTransform, WriteBatch};
 use serde;
 use tempfile::{tempdir, TempDir};
 
@@ -34,17 +34,8 @@ struct PersistentMeta {
 
 #[derive(Clone)]
 struct PersistentIndex {
-    column_family: ColumnFamily,
+    column_family: String,
     columns: Vec<usize>,
-}
-
-impl PersistentIndex {
-    fn new(column_family: ColumnFamily, columns: Vec<usize>) -> Self {
-        Self {
-            column_family,
-            columns,
-        }
-    }
 }
 
 /// PersistentState stores data in RocksDB.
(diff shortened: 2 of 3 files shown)


noria-server/dataflow/src/state/persistent_state.rs
```diff
@@ -206,31 +128,32 @@ impl State for PersistentState {
...
-            .as_mut()
-            .unwrap()
-            .create_cf(&index_id, &self.db_opts)
-            .unwrap();
+        let db = self.db.as_mut().unwrap();
+        db.create_cf(&index_id, &self.db_opts).unwrap();
 
         // Build the new index for existing values:
         if !self.indices.is_empty() {
-            for chunk in self.all_rows().chunks(INDEX_BATCH_SIZE).into_iter() {
```
ekmartin: Not a huge deal, but I couldn't get my `all_rows()` helper to work here: 

```rust
error[E0502]: cannot borrow `*self` as immutable because it is also borrowed as mutable
   --> noria-server/dataflow/src/state/persistent_state.rs:139:26
    |
131 |         let column_family = self.db
    |                             ------- mutable borrow occurs here
...
139 |             for chunk in self.all_rows().chunks(INDEX_BATCH_SIZE).into_iter() {
    |                          ^^^^ immutable borrow occurs here
...
145 |                     batch.put_cf(column_family, &key, value).unwrap();
    |                                  ------------- mutable borrow later used here

```

noria-server/dataflow/src/state/persistent_state.rs
```diff
@@ -511,9 +439,9 @@ impl PersistentState {
     }
 
     // Filters out secondary indices to return an iterator for the actual key-value pairs.
-    fn all_rows(&self) -> impl Iterator<Item = (Box<[u8]>, Box<[u8]>)> {
+    fn all_rows(&self) -> impl Iterator<Item = (Box<[u8]>, Box<[u8]>)> + '_ {
```
ekmartin: The compiler suggested that I added `'_` here but I have no idea if that's a sane thing to do—here's the error:

```rust
error: lifetime may not live long enough
   --> noria-server/dataflow/src/state/persistent_state.rs:442:27
    |
442 |     fn all_rows(&self) -> impl Iterator<Item = (Box<[u8]>, Box<[u8]>)> {
    |                 -         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ opaque type requires that `'1` must outlive `'static`
    |                 |
    |                 let's call the lifetime of this reference `'1`
help: to allow this impl Trait to capture borrowed data with lifetime `'1`, add `'_` as a constraint
    |
442 |     fn all_rows(&self) -> impl Iterator<Item = (Box<[u8]>, Box<[u8]>)> + '_ {
```

jonhoo: Oh, yeah, that's probably fine --- `impl Trait` by default assumes the lifetime of all input arguments, so `'_` is a way of indicating that it shouldn't capture the lifetime of `self`.

noria-server/dataflow/src/state/persistent_state.rs
jonhoo: Ah, this is an artifact of you now setting
```rust
let db = self.db.as_mut().unwrap();
```
And then using `db` repeatedly below, instead of using `self.db.get_mut()` repeatedly below. When you call `self`, the compiler detects that `self` is already borrowed in `db`, and so it disallows the call.

jonhoo:
This looks good! Out of curiosity, do you think the repeated lock acquisition is going to be a performance bottleneck here at all?

ms705:
Changes look good to me -- might be good to get this merged before things diverge too much.

Is there anything that needs doing before?

noria-server/dataflow/Cargo.toml
```diff
@@ -35,6 +35,7 @@ common = { path = "../common" }
 noria = { path = "../../noria" }
 
 [dependencies.rocksdb]
-git = "https://github.com/ekmartin/rust-rocksdb.git"
+git = "https://github.com/rust-rocksdb/rust-rocksdb.git"
+default-features = false
 features = ["lz4"]
-branch = "custom"
+rev = "184440b2095979baeb747ccf300022198e9606ed"
```
jonhoo: Is there a particular reason you leave this as a `git` dependency? Isn't this all in `0.12`?

jonhoo:
I think the comment I just left is the only real question I have. Some of the code may be possible to simplify a little, but I don't think it's very important.

jonhoo:
Actually, let's just merge this and then I'll try just moving to a regular dependency.

ekmartin:
Sorry about my unresponsiveness here, and thanks for fixing the git dependency (12.2 wasn't released when I opened this PR).

PR #127, opened 2019-07-03
Factored and simplified a function out of `on_input`

JustusAdam:
> A small change for a module, a big change for humanity

... Actually no 😄 , but I stumbled upon this section when trying to understand the code and I think this version is more readable. It doesn't break any tests and, according to taster doesn't appear to cause slowdown.

### DIFF
--- a/noria-server/dataflow/src/ops/grouped/mod.rs
+++ b/noria-server/dataflow/src/ops/grouped/mod.rs
@@ -98,6 +98,12 @@ impl<T: GroupedOperation> GroupedOperator<T> {
 
 /// Extract a copy of all values in the record being targeted by the group
 fn get_group_values(group_by: &[usize], row: &Record) -> Vec<DataType> {
+    // This attribute is only here, because `is_sorted` is unstable. I didn't
+    // want to add a `feature` to the crate for a debug assertion thus I guarded
+    // both the `feature` and this assertion are with `cfg(debug)` (see also
+    // `lib.rs`)
+    #[cfg(debug)]
+    debug_assert!(group_by.is_sorted());
     let mut group = Vec::with_capacity(group_by.len() + 1);
     for &group_idx in group_by {
         group.push(row[group_idx].clone())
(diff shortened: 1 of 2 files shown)


noria-server/dataflow/src/ops/grouped/mod.rs
```diff
@@ -96,6 +96,16 @@ impl<T: GroupedOperation> GroupedOperator<T> {
     }
 }
 
+/// Extract a copy of all values in the record being targeted by the group
+fn get_group_values(group_by: &[usize], row: &Record) -> Vec<DataType> {
+    let mut group = Vec::with_capacity(group_by.len() + 1);
+    for group_idx in group_by.iter() {
```
jonhoo:
```suggestion
    for &group_idx in group_by {
```

jonhoo: With that change, you can also remove the `*` on the line below.

noria-server/dataflow/src/ops/grouped/mod.rs
```diff
@@ -96,6 +96,16 @@ impl<T: GroupedOperation> GroupedOperator<T> {
...
+        group.push(row[*group_idx].clone())
+    }
+    debug_assert!(group.len() == group_by.len());
```
jonhoo:
```suggestion
    debug_assert_eq!(group.len(), group_by.len());
```

jonhoo:
This looks like a good change! I think there _is_ a difference here in how the two code segments would work if `group_by` is not in sorted order, but I think the old code was also just broken for that use-case, so this is probably fine :) Some minor nits above, but apart from that :+1:

JustusAdam:
Yeah, i it seems the old code would omit out-of-order columns, while this one should emit a filtered row in the same order as the `group_by`.
Not really sure if either is better in that case. The old one would most likely fail faster.
Perhaps I could add a `debug_assert!` checking that the columns are sorted? That might make it more reliable.
I just find this version to me much easier to read.

jonhoo:
I agree! Yeah, that seems like a reasonable assertion to add. Happy to merge after that.

jonhoo:
We should also probably fix Travis again :sweat_smile:

JustusAdam:
Apparently travis actually worked there for once.... Maybe don't make the compilation process silent (should that be the case)? Travis likes you to do lots of output while building 😄

jonhoo:
I think I'd prefer if you just asserted that each column is numerically greater than the previous one. Just keep `last_col: usize` and check + re-assign on each iteration. You shouldn't need `is_sorted`.

jonhoo:
Oh, @ms705 already merged it.. :sweat_smile: I'd prefer not to have conditional features, as it makes it even harder to get off of nightly.

PR #128, opened 2019-07-05
Refactored state impl via macro and trait

JustusAdam:
The idea here is to make the conversion between `Row` (or `&[DataType]`) and an
index for the state map reusable, namely for other `State` impls.
It just so happens that that also allows me to apply some do-not-repeat-yourself
principles to the state impls using small `macro_rules`.

I did this change in preparation of implementing a custom `State`. I don't necessarily mean to suggest that this version is *better*, and *should* be merged` but I thought I'd inform you about the possibility.

The code isn't actually shorter, which I think is a bit unfortunate, but perhaps more readable.

### DIFF
--- a/noria-server/dataflow/src/state/single_state.rs
+++ b/noria-server/dataflow/src/state/single_state.rs
@@ -1,3 +1,4 @@
+use super::mk_key::MakeKey;
 use common::SizeOf;
 use prelude::*;
 use rand::prelude::*;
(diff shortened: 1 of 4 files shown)


noria-server/dataflow/src/state/single_state.rs
```diff
@@ -10,6 +11,28 @@ pub(super) struct SingleState {
     partial: bool,
     rows: usize,
 }
+
+macro_rules! insert_row_match_impl {
+    ($self:expr, $r:expr, $map:expr) => {{
```
jonhoo: These should probably all be `:ident`.

noria-server/dataflow/src/state/single_state.rs
```diff
@@ -10,6 +11,28 @@ pub(super) struct SingleState {
...
+        match $map.entry(key) {
+            Entry::Occupied(mut rs) => rs.get_mut().push($r),
+            Entry::Vacant(..) if $self.partial => return false,
+            rs @ Entry::Vacant(..) => rs.or_default().push($r),
+        }
+    }};
+}
+
+macro_rules! remove_row_match_impl {
+    ($self:expr, $r:expr, $do_remove:expr, $map:expr) => {{
```
jonhoo: These should probably also all be `:ident`.

jonhoo:
This seems like a reasonable change to me! I'd be happy to merge with the two minor macro changes above.

PR #139, opened 2019-10-06
Persistent state timestamp

zeling:
- Added basic versioning to the persistent_state
- Should come up with a way to propagate versions to other parts of the system in later commits.


ms705:
Looking good! I'll merge this into the `mvcc` branch.

PR #140, opened 2019-10-11
Disable jemalloc on windows to allow windows build

DoumanAsh:
This is work-around to enable `async-await` branch to be built on windows.

Proper fix would require fixing jemalloc build script (which is hard for msvc as jemalloc uses autoconf)

### DIFF
--- a/noria-server/dataflow/src/lib.rs
+++ b/noria-server/dataflow/src/lib.rs
@@ -8,6 +8,7 @@
 #![deny(unused_extern_crates)]
 #![allow(clippy::redundant_closure)]
 
+#[cfg(not(target_env="msvc"))]
 #[global_allocator]
 static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,12 +19,21 @@ stages:
             vmImage: ubuntu-16.04
           MacOS:
             vmImage: macOS-10.14
+          Windows:
+            vmImage: windows-2019
       pool:
         vmImage: $(vmImage)
       steps:
        - template: azure/install-rust.yml@templates
          parameters:
            rust: nightly
+       - powershell: |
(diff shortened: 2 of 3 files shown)


ms705:
Thanks for this! To confirm: with this change, the `async-await` branch builds and runs on Windows?

@jonhoo Seems like a worthwhile change to merge, no? Performance without jemalloc could be underwhelming (based on our previous measurements), but making it possible to at least develop on Windows seems reasonable.

noria-server/dataflow/src/lib.rs
```diff
@@ -8,6 +8,11 @@
 #![deny(unused_extern_crates)]
 #![allow(clippy::redundant_closure)]
 
+#[cfg(target_env="msvc")]
+#[global_allocator]
+static ALLOC: std::alloc::System = std::alloc::System;
```
jonhoo: Do we even need to specify this? Isn't it the default?

DoumanAsh: Actually yeah, it is not no_std build so default allocator should be enabled already.

Let me check and re-push

DoumanAsh:
>Thanks for this! To confirm: with this change, the async-await branch builds and runs on Windows?

Yes, the only requirement is to add `LIBCLANG_PATH` for bindgen in rocksbd

noria-server/dataflow/src/lib.rs
DoumanAsh: Updated PR

jonhoo:
Actually, would you mind adding Windows to the testing matrix too?
https://github.com/mit-pdos/noria/blob/6c1bfd1e5aab0afee659e03925babbecca44faf9/azure-pipelines.yml#L17-L21

I think:
```yaml
Windows:
  vmImage: windows-2019
```
is what you want

DoumanAsh:
Is azure CI not enabled for PR checks?

jonhoo:
Huh, apparently not. Should be fixed now if you try to push again.

jonhoo:
Tests fail, but that's a separate issue. Let's merge.

PR #150, opened 2019-11-23
Filtered aggregate

jmbredenberg:
Automatically merges adjacent filters and aggregations into a single operator when possible, so that a single traversal of the data performs both functions. Goes with the nom-sql commit adding syntax for COUNT/SUM(CASE WHEN x THEN y ELSE z END).

Adds various tests for the operator and for the automatic merging. Also refactors the way we store condition vectors, and adds support for generating such a vector for the AND of two condition trees.


Cargo.toml
```diff
@@ -8,3 +8,6 @@ members = [
 
 [profile.release]
 debug=true
+
+[patch.crates-io]
+nom-sql = { path = "../nom-sql" }
```
jonhoo: This is probably not something we want to commit :)

noria-benchmarks/lobsters/src/main.rs
```diff
@@ -15,12 +15,16 @@ use trawler::{LobstersRequest, UserId};
 const ORIGINAL_SCHEMA: &'static str = include_str!("../db-schema/original.sql");
 const NORIA_SCHEMA: &'static str = include_str!("../db-schema/noria.sql");
 const NATURAL_SCHEMA: &'static str = include_str!("../db-schema/natural.sql");
+const CAN_COMBINE_SCHEMA: &'static str = include_str!("../db-schema/cancombine.sql");
+const FILTER_AGGREGATE_SCHEMA: &'static str = include_str!("../db-schema/filteraggregate.sql");
```
jonhoo: How do these differ from the other ones?

noria-server/dataflow/src/ops/filter.rs
```diff
@@ -10,7 +10,7 @@ use prelude::*;
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Filter {
     src: IndexPair,
-    filter: sync::Arc<Vec<Option<FilterCondition>>>,
+    filter: sync::Arc<Vec<(usize, FilterCondition)>>,
```
jonhoo: What's the reason for this change? It'll incur some performance overhead if I'm reading the change right?

noria-server/dataflow/src/ops/mod.rs
```diff
@@ -21,6 +21,7 @@ pub enum NodeOperator {
     Sum(grouped::GroupedOperator<grouped::aggregate::Aggregator>),
     Extremum(grouped::GroupedOperator<grouped::extremum::ExtremumOperator>),
     Concat(grouped::GroupedOperator<grouped::concat::GroupConcat>),
+    FilterSum(grouped::GroupedOperator<grouped::filteraggregate::FilterAggregator>),
```
jonhoo: It's a little weird for this to be called `FilterSum` when the inner type is `FilterAggregator` and supports both sum and count. Could this just be `Filtered` maybe?

noria-server/mir/src/node.rs
```diff
@@ -402,7 +420,15 @@ pub enum MirNodeType {
     },
     /// filter conditions (one for each parent column)
     Filter {
-        conditions: Vec<Option<FilterCondition>>,
+        conditions: Vec<(usize, FilterCondition)>,
```
jonhoo: Same here -- why is this change necessary?

noria-server/mir/src/node.rs
```diff
@@ -485,8 +511,9 @@ impl MirNodeType {
             } => {
                 group_by.push(c);
             }
-            MirNodeType::Filter { ref mut conditions } => {
-                conditions.push(None);
+            MirNodeType::Filter { .. } => {}
```
jonhoo: Is it intentional that this no longer does a `conditions.push`?

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+use dataflow::ops::grouped::filteraggregate::FilterAggregation;
+use dataflow::ops::grouped::aggregate::Aggregation;
 use node::{MirNode, MirNodeType};
 use query::MirQuery;
 use MirNodeRef;
 
 use std::collections::HashMap;
 
-pub fn optimize(q: MirQuery) -> MirQuery {
+pub fn optimize(mut q: &mut MirQuery) -> Vec<MirNodeRef> {
```
jonhoo: It'd be good to document what this returns.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                    let child = temp.children.first().unwrap().borrow();
+                    match child.inner {
+                        MirNodeType::Filter { ref conditions } => {
+                            if child.ancestors.len() == 1 {
+                                candidate = true;
+
+                                // But wait -- need to check if the filter is on the aggregation result
+                                use nom_sql::FunctionExpression::{Count, Sum};
+                                for (i, col) in child.columns.iter().enumerate() {
+                                    match col.function.clone() {
```
jonhoo: This can be `if let Some(func_expr) = ` to save a level of indentation.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                    };
+                }
+            },
+            MirNodeType::Aggregation { ref on, .. }  => {
+                // if there's exactly one child and it's an aggregation and it has exactly one parent,
+                // then this is a candidate
+                if n.borrow().children.len() == 1 {
+                    let temp = n.borrow();
+                    let child = temp.children.first().unwrap().borrow();
+                    match child.inner {
```
jonhoo: This can be `if let MirNodeType::Filter { ref conditions } = child.inner` to save a level of indentation

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+        // now scan for candidacy
+        let mut candidate = false;
+        match n.borrow().inner {
+            MirNodeType::Filter { .. }  => {
+                // if there's exactly one child and it's an aggregation and it has exactly one parent,
+                // then this is a candidate
+                if n.borrow().children.len() == 1 {
+                    let temp = n.borrow();
+                    let child = temp.children.first().unwrap().borrow();
+                    match child.inner {
```
jonhoo: This can be `if let MirNodeType::Aggregation { .. } = child.inner` to save a level of indentation

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+
+                                // But wait -- need to check if the filter is on the aggregation result
+                                use nom_sql::FunctionExpression::{Count, Sum};
+                                for (i, col) in child.columns.iter().enumerate() {
+                                    match col.function.clone() {
+                                        Some(func_expr) => {
+                                            match *func_expr {
+                                                 Count(col, _)
+                                                 | Sum(col, _) => {
+                                                    if col.name == on.name {
```
jonhoo: This can be a match guard instead (just move it before the `=>`) to save another level of indentation.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                                                        // this column may be the aggregation result
+                                                        // so if we're filtering on it, we're not a candidate
+                                                        for (j, _cond) in conditions {
+                                                            if *j == i {
+                                                                candidate = false;
```
jonhoo: Do we want to break early if this is the case to avoid looking at a bunch of other conditions/children?

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                            }
+                        },
+                        _ => {},
+                    };
+                }
+            },
+            MirNodeType::Aggregation { ref on, .. }  => {
+                // if there's exactly one child and it's an aggregation and it has exactly one parent,
+                // then this is a candidate
+                if n.borrow().children.len() == 1 {
```
jonhoo: I would probably invert this:
```rust
if n.borrow().children.len() !=1 {
    continue;
}
```
To save a level of indentation.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                    let temp = n.borrow();
+                    let child = temp.children.first().unwrap().borrow();
+                    match child.inner {
+                        MirNodeType::Filter { ref conditions } => {
+                            if child.ancestors.len() == 1 {
```
jonhoo: I would probably invert this as well to save another level of indentation.

jonhoo: Also, I think this `if` could be moved to outside the `match`?

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+        if !visited_nodes.get(&node_name).unwrap() {
+            continue;
+        }
+        // now scan for candidacy
+        let mut candidate = false;
+        match n.borrow().inner {
+            MirNodeType::Filter { .. }  => {
+                // if there's exactly one child and it's an aggregation and it has exactly one parent,
+                // then this is a candidate
+                if n.borrow().children.len() == 1 {
```
jonhoo: Since this `if` is the same for both match arms, maybe just move it outside the match entirely?

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+        let mut new_name = child.name.clone();
+        new_name.push_str("_filteragg");
+
+        let new_node = MirNode::new(
+            &new_name,
+            child.from_version,
+            child.columns.clone(),
+            MirNodeType::FilterAggregation {
+                on: match &agg {
+                    MirNodeType::Aggregation { on, group_by: _, kind: _ } => on.clone(),
```
jonhoo: You can just use `{ on, .. }` here :)

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                    _ => unreachable!(),
+                },
+                else_on: None,
+                group_by: match &agg {
+                    MirNodeType::Aggregation { on: _, group_by, kind: _ } => group_by.to_vec(),
```
jonhoo: And `{ group_by, .. }` here.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+
+        let mut new_name = child.name.clone();
+        new_name.push_str("_filteragg");
+
+        let new_node = MirNode::new(
+            &new_name,
+            child.from_version,
+            child.columns.clone(),
+            MirNodeType::FilterAggregation {
+                on: match &agg {
```
jonhoo: This should probably be a `if let`.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                    MirNodeType::Aggregation { on, group_by: _, kind: _ } => on.clone(),
+                    _ => unreachable!(),
+                },
+                else_on: None,
+                group_by: match &agg {
```
jonhoo: This should probably be a `if let`.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                    MirNodeType::Aggregation { on: _, group_by, kind: _ } => group_by.to_vec(),
+                    _ => unreachable!(),
+                },
+                //group_by.into_iter().cloned().collect(),
+                kind: match &agg {
```
jonhoo: This should probably be an `if let`.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                    MirNodeType::Aggregation { on: _, group_by: _, kind } => {
+                        match kind {
+                            Aggregation::COUNT => FilterAggregation::COUNT,
+                            Aggregation::SUM => FilterAggregation::SUM,
+                        }
+                    },
+                    _ => unreachable!(),
+                },
+                conditions: match &filter {
```
jonhoo: This should probably be an `if let`.

noria-server/dataflow/src/ops/filter.rs
ms705: It allows merging ANDed filters over the same columns (e.g. `x < 5 and x > 3`) into a single filter (two entries for the same column index).

Why would it incur performance overhead? Previously, we had to iterate over lots of `None` entries in the vector, and additionally create the columns indexes via `enumerate`, all on the critical path. Seems to me like this should at least be a neutral change from a performance PoV.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+                    MirNodeType::Filter { ref conditions } => conditions.to_vec(),
```
jonhoo: Hmm, I wonder if `conditions` needs some adjustment here depending on whether the original filter was before or after given that the aggregation moves columns around?

noria-server/mir/src/node.rs
jmbredenberg: Yep! The new representation of conditions only adds an element of (index, condition) if the condition is present, instead of having either Some(condition) or None for every index. See Malte's note above about why I changed the representation.

I read your comments -- thank you so much for the incredibly detailed feedback! I haven't used Rust enough to be familiar with the conventions, so it's very helpful -- but I probably won't get a chance to make the changes until after Weds.

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,217 @@
...
+            },
+            n.borrow().ancestors.clone(),
+            child.children.clone(),
+        );
+
+        // now update parents/children to reference the new node
+        for c in child.children.iter() {
+            let mut new_ancestors = Vec::new();
+            for a in c.borrow().ancestors.iter() {
+                // TODO is versioned_name sufficiently unique for here (and below)?
```
jonhoo: @ms705

noria-server/mir/src/query.rs
```diff
@@ -71,9 +71,10 @@ impl MirQuery {
         mut self,
         table_mapping: Option<&HashMap<(String, Option<String>), String>>,
         sec: bool,
-    ) -> MirQuery {
+    ) -> (MirQuery, Vec<MirNodeRef>) {
```
jonhoo: It'd be good to document the return type here as well.

noria-server/mir/src/visualize.rs
```diff
@@ -149,6 +150,24 @@ impl GraphViz for MirNodeType {
                     .join(", ");
                 write!(out, "{} | γ: {}", op_string, group_cols)?;
             }
+            MirNodeType::FilterAggregation {
+                ref on,
+                else_on: _,
+                ref group_by,
+                ref kind,
+                conditions: _,
```
jonhoo: You don't need all of these `: _`. Just list the fields you want, and then add `, ..`.

noria-server/src/controller/sql/mir/mod.rs
```diff
@@ -153,14 +153,52 @@ impl SqlToMirConverter {
             })
     }
 
+    // TODO: remove (this is just for debugging)
```
jonhoo: Should this now be removed?

noria-server/src/controller/sql/mir/mod.rs
```diff
@@ -936,12 +975,13 @@ impl SqlToMirConverter {
...
     ) -> Vec<MirNodeRef> {
         use dataflow::ops::grouped::aggregate::Aggregation;
+        use dataflow::ops::grouped::filteraggregate::FilterAggregation;
         use dataflow::ops::grouped::extremum::Extremum;
         use nom_sql::FunctionExpression::*;
 
         let mut out_nodes = Vec::new();
 
-        let mknode = |over: &Column, t: GroupedNodeType, distinct: bool| {
+        let mknode = |over: &Column, over_else: Option<Literal>, t: GroupedNodeType, distinct: bool, cond: Option<&ConditionExpression>| {
```
jonhoo: I think the compiler _might_ now let you get away with not giving the types for all these closure arguments.

noria-server/src/controller/sql/mir/mod.rs
```diff
@@ -1056,6 +1124,34 @@ impl SqlToMirConverter {
                 vec![parent_node.clone()],
                 vec![],
             ),
+            GroupedNodeType::FilterAggregation(filter_agg) => {
+                use nom_sql::ConditionExpression::*;
+
+                let cond = condition.expect("Filter aggregation must have condition!");
```
jonhoo: Given that this is the case, it might be better to just stick the conditions inside the `FIlterAggregation` variant rather than passing it separately?

noria-server/src/integration.rs
```diff
@@ -2214,6 +2214,7 @@ fn test_queries(test: &str, file: &'static str, shard: bool, reuse: bool, log: b
             }
         }
     });
+    println!("{:#?}", g.graphviz());
```
jonhoo: This was probably left in by accident?

noria-server/src/pdfify.sh
```diff
@@ -0,0 +1,19 @@
+FILE=$1
```
jonhoo: What does this script do?

noria-server/dataflow/src/ops/filter.rs
jonhoo: Previously, given a single input index, you could efficiently determine the corresponding output index. Maybe the inverse lookup is more common though, in which case this is a perf win. I don't know :)

jonhoo:
@jmbredenberg 
> I read your comments -- thank you so much for the incredibly detailed feedback! I haven't used Rust enough to be familiar with the conventions, so it's very helpful -- but I probably won't get a chance to make the changes until after Weds.

Happy to! I hope it doesn't come across as too overwhelming. This is good work. I just wanted to make sure we didn't have anything slip through the cracks :) No rush in making the changes. Also, separately, I highly recommend `cargo clippy` — it's great at pointing out non-idiomatic Rust code!

noria-benchmarks/lobsters/src/main.rs
jmbredenberg: Originally I thought this was how I would go about testing things and making graphs. Then Malte pointed me at adding a test to integration.rs, and I forgot to remove the added schema here. I'll do that now.

noria-server/dataflow/src/ops/filter.rs
jmbredenberg: I remember changing a lot of places where we iterated over the entire vector. I don't remember whether there are any where we needed to lookup by index.

But yeah, as Malte said, the goal was to make it possible to handle slightly more complex conditions.

noria-server/dataflow/src/ops/mod.rs
jmbredenberg: Yeah, I agree with that. I don't fully understand what this file is doing. I was just trying to be consistent with the above, where it defines Sum on Aggregator even though that also supports sum and count. Can you shed any light on why it's doing that?

noria-server/mir/src/optimize.rs
jmbredenberg: Agreed. Done.

noria-server/mir/src/optimize.rs
jmbredenberg: Sweet! I wasn't familiar with this pattern, so thanks for flagging it.

noria-server/mir/src/optimize.rs
jmbredenberg: Same here. This looks a lot less horrendously over-nested now :)

noria-server/mir/src/optimize.rs
jmbredenberg: Same here. This looks a lot less horrendously over-nested now :)

noria-server/mir/src/optimize.rs
jmbredenberg: Sounds good.

noria-server/mir/src/optimize.rs
jmbredenberg: Done.

noria-server/mir/src/optimize.rs
jmbredenberg: Done, done.

noria-server/mir/src/optimize.rs
jmbredenberg: Done.

noria-server/mir/src/optimize.rs
jmbredenberg: So, in this case I did know there was a shorter way to write it, but I personally found it more readable/modifiable to see what all the fields are and which ones we're using here. Why do you prefer the other way?

I'll defer to your preferred style here, of course. I'm just curious for your reasoning.

noria-server/mir/src/optimize.rs
jmbredenberg: Hmm, I do like having the unreachable() to cause an assertion failure if it ever does end up hitting that code path. Does it work if I just use "let" instead of "if let"? 

(Mostly musing out loud; I can check for myself whether that compiles.)

noria-server/mir/src/optimize.rs
jmbredenberg: Great point. Will think about this more soon.

noria-server/mir/src/query.rs
jmbredenberg: Done.

noria-server/mir/src/visualize.rs
jmbredenberg: Done.

noria-server/src/controller/sql/mir/mod.rs
jmbredenberg: Good call. Sorry about that -- I was in a rush last week and forgot to clean up things like this.

noria-server/src/controller/sql/mir/mod.rs
jmbredenberg: As someone not yet very familiar with the codebase, I find it helpful to know what all the types are when I read things like this.

noria-server/src/controller/sql/mir/mod.rs
jmbredenberg: I spent a while trying to do this, and it's actually a huge pain, primarily because mir_to_flow also wants to pass conditions along with the GroupedNodeType but in that case the conditions are Vec<(usize, FilterCondition)>s rather than ConditionExpressions. So then I tried just making them both Vec<(usize, FilterCondition)>s and just pushing the logic to convert from ConditionExpressions earlier in mir/mod.rs. But this snowballed into pushing a lot of logic earlier because to_conditions requires already having made the parent node...

I'm going to give up on this one, though I agree it would've been a lot more elegant.

noria-server/src/integration.rs
jmbredenberg: Er, yeah. That's how I was printing out a graph to check whether my optimization was actually doing anything on the real queries.

noria-server/src/pdfify.sh
jmbredenberg: Oh boy, I forgot about that. After I did the above println it had a bunch of extra backslashes and such that were preventing me from being able to turn it into a pdf, so I made a script to clean it up and automatically make the graph out of it. I wanted it in version control so I wouldn't lose it, but I'll take it out now that it's time to submit.

Cargo.toml
jmbredenberg: Yup! I'll do that last; I want it to still work while I'm testing it :)

noria-server/mir/src/optimize.rs
jmbredenberg: It looks like the reason this was working before was because when merging {parent, child} I was giving the new node child.columns for its column vec. I think instead it should be getting parent.columns and adjusting the filter indices as you suggested.

...Actually, now I'm blanking. Dumb question: are the columns the input or output columns of the node? (Either way I think it needs the fix you suggested.)

jonhoo: I believe columns is the output, but I don't know what MIR does here, so we'd have to check with @ms705.

noria-server/mir/src/optimize.rs
jonhoo: Sure!
```rust
if let MirNodeType::Aggregation { ... } = agg {
    // ...
} else {
    unreachable!()
}
```

noria-server/mir/src/optimize.rs
jonhoo: Ah, so, the reason is in case we later add _more_ fields to `Aggregation`. This pattern would then fail to compile, whereas one with `, ..` would not. Although that could also be considered a feature, depending on how sensitive you think this could would be to `Aggregation` changing.

noria-server/dataflow/src/ops/mod.rs
jonhoo: Ah, so, the type here is just to have a single type that can contain all operators. That's necessary when we, for example, need a type for the data-flow graph as a hole.

jonhoo:
As an aside, as you fix things, feel free to mark comments that have been addressed fully as "resolved" in the GitHub UI :)

noria-server/mir/src/optimize.rs
jmbredenberg: Yup, just confirmed that with Malte. Will finish fixing this now.

noria-server/mir/src/optimize.rs
jmbredenberg: Yeah, I ended up with that, but I wouldn't call it easily readable :P

noria-server/dataflow/src/ops/mod.rs
jmbredenberg: OK. I'm still not understanding why one of its enum options is Sum instead of Aggregation?

noria-server/mir/src/optimize.rs
jmbredenberg: OK, I have written some code that I believe will catch this case.

I believe previously all the tests were passing because they were Aggregation followed by Filter, and the column reordering would only cause a problem with Filter followed by Aggregation. I am having trouble coming up with a simple test case that forces it to be in that order. The obvious way to write a query that does this would be to aggregate on the column being filtered, because then you have to do the filter first, but then I realized this case can't merge the two operators anyway because the filter column doesn't exist in the output columns and thus there's no way to represent it. (This is an interesting limitation of our current filter model.)

Anyway, now there's *also* code to make sure we're not merging in that case, and a test for that, but no test for the original problem you spotted. Feedback on the new code would be much appreciated.

jmbredenberg: Never mind, needs another rewrite -- Malte just pointed out that, although the columns field is the output, the filter indices are over the input (i.e. the parent's columns field). 

Oh, well -- at least I understand the code base better now.

jonhoo: It is probably also okay for the code to just _detect_ this case and error, rather than coming up with the correct implementation for it, since it seems to be tricky to trigger.

noria-server/mir/src/optimize.rs
```diff
@@ -57,64 +58,56 @@ fn find_and_merge_filter_aggregates(q: &mut MirQuery) -> Vec<MirNodeRef> {
...
+                if let MirNodeType::Filter { ref conditions } = child.inner {
+                    if child.ancestors.len() != 1 {
+                        continue;
+                    }
+                    candidate = true;
+
+                    // But wait -- need to check if the filter is on the aggregation result
+                    use nom_sql::FunctionExpression::{Count, Sum};
+                    for (i, col) in child.columns.iter().enumerate() {
+                        if let Some(func_expr) = col.function.clone() {
```
jonhoo: I don't _think_ this needs to `.clone`(). Can't it just be:
```rust
match col.function {
  Some(Count(ref col, _)) | Some(Sum(ref col, _)) if col.name == on.name => {
```

noria-server/mir/src/optimize.rs
```diff
@@ -149,29 +140,23 @@ fn find_and_merge_filter_aggregates(q: &mut MirQuery) -> Vec<MirNodeRef> {
             child.from_version,
             child.columns.clone(),
             MirNodeType::FilterAggregation {
-                on: match &agg {
-                    MirNodeType::Aggregation { on, group_by: _, kind: _ } => on.clone(),
-                    _ => unreachable!(),
-                },
+                on: if let MirNodeType::Aggregation { on, .. } = &agg {
```
jonhoo: Looking at this again, maybe the `if let` should just go on the outside of this constructor instead? Then you can match on all the fields at once, and just have one `unimplemented!` clause.

noria-server/src/controller/sql/mir/mod.rs
```diff
@@ -199,25 +233,34 @@ impl SqlToMirConverter {
...
 
         match columns.iter().rposition(|c| *c.name == l.name) {
             None => {
                 // Might occur if the column doesn't exist in the parent; e.g., for aggregations.
                 // We assume that the column is appended at the end.
-                columns.push(Column::from(l));
-                filters.push(f);
+                match n.borrow().inner {
+                    MirNodeType::Aggregation { .. }  => {
+                        columns.insert(columns.len()-1, Column::from(l));
```
ms705: Please update comment above to say that we append the column at the end unless we have and aggregation, in which case it needs to go before the computed column, which is last.

noria-server/src/controller/sql/mir/mod.rs
```diff
@@ -952,34 +997,62 @@ impl SqlToMirConverter {
...
+            ),
+            Sum(FunctionArguments::Conditional(CaseWhenExpression{
+                ref condition,
+                then_expr: ColumnOrLiteral::Column(ref col),
+                else_expr: Some(ColumnOrLiteral::Literal(ref else_val))
+            }), _) => mknode(
+                &Column::from(col),
+                Some(else_val.clone()),
+                GroupedNodeType::FilterAggregation(FilterAggregation::SUM),
+                false,
```
ms705: Should this code assert that `distinct` (`_` above) is false?

noria-server/src/controller/sql/mod.rs
```diff
@@ -1002,16 +1006,21 @@ mod tests {
...
+        let mut a_vec: Vec<&Column> = attrs.to_vec();
+        a_vec.sort();  // QueryGraph.signature() sorts them, so we must to match
+        for a in &a_vec {
             a.hash(&mut hasher);
         }
         for c in columns.iter() {
             OutputColumn::Data((*c).clone()).hash(&mut hasher);
         }
-        hasher.finish()
+        let temp = hasher.finish();
```
ms705: Is this indirection via a temporary needed?

thomasharmon808:
Are there any things left that need finishing up that I could help with? I saw @ms705 's approval for the PR but still some open comments and am willing to help fix up any last issues

jmbredenberg:
> Are there any things left that need finishing up that I could help with? I saw @ms705 's approval for the PR but still some open comments and am willing to help fix up any last issues

Thanks for the offer! I was working on something else and got a bit distracted from finishing this up, but it should be done soon.

noria-server/src/controller/sql/mod.rs
jmbredenberg: Oops, I think I added that for printing purposes and then forgot to fully remove it when I deleted the print statement. Fixed.

noria-server/mir/src/optimize.rs
jmbredenberg: This seems to fail because the FunctionExpression is boxed. Is there a way to write something like Some(Box(Count(...)))? I don't know much about boxes.

jonhoo: Ah, yes, that's a little awkward. Easiest way is
```rust
match col.function.as_deref() {
```

noria-server/mir/src/node.rs
```diff
@@ -375,12 +391,14 @@ impl MirNode {
 
 /// Specifies the adapatation of an existing base node by column addition/removal.
 /// `over` is a `MirNode` of type `Base`.
+#[derive(Clone)]
```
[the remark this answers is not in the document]
jmbredenberg: It's needed in order to #[derive(Clone)] on MirNodeType, which I am using at optimize.rs:160. (If I don't clone there, I get "error[E0507]: cannot move out of dereference of `std::cell::Ref<'_, node::MirNode>`")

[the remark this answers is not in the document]
ms705: I see! If I read that code correctly, it's trying to maintain a reference to the filter condition and the associated aggregation. One way you could do this without having to clone the aggregation's `MirNodeType` in `inner` is to just pass around a `MirNodeRef` to the aggregation's node (`MirNodeRef` is already `Clone`).

noria-server/mir/src/optimize.rs
```diff
@@ -1,19 +1,235 @@
...
+                    // for each column in filter's conditions, we need to figure out
+                    // which column in input to aggregation has the same name so we
+                    // can reindex the condition to that column
+                    let mut conds = Vec::new();
+                    for (i, cond) in conditions {
+                        let mut found = 0;
+                        // conditions are on the parent columns of the filter node, i.e. agg's columns
+                        let cond_col = n.borrow().columns[*i].clone();
+                        // we need to reindex them onto the new filteragg's parent columns, i.e. agg's parent's columns
+                        assert!(n.borrow().ancestors.len() == 1);
```
jonhoo: A handy macro for these types of comparisons is `assert_eq!`. It does the same thing, but if the values are _not_ equal, it will print what the values were to help with debugging. There is also `assert_ne!` for the reverse case.

noria-server/mir/src/optimize.rs
jonhoo: What was the final decision on this `TODO`?

noria-server/src/integration.rs
```diff
@@ -2476,6 +2476,30 @@ async fn soupy_lobsters() {
     .await;
 }
 
+#[tokio::test(threaded_scheduler)]
+async fn mergeable_lobsters() {
```
jonhoo: How do the schemas here differ from the regular lobsters schemas?

PR #152, opened 2020-01-25
Use a type alias for FirstByte instead of boxing

udoprog:
While rust-lang/rust#64445 is still a thing, we can use a type alias in a different way to make it work temporarily.


jonhoo:
Ah, that's a great solution, thank you!

PR #169, opened 2020-09-03
Fix link to controller was broken with directory changes

ptravers:
Not totally sure if this is the same as the original linked code. But it seemed the most similar of the code I could see.


ms705:
Thanks for this! The fix for `data.rs` looks good, but the equivalent path to `inner.rs` is now `server/src/controller/inner.rs`.

ptravers:
@ms705 ah cool! will fix that now!

ptravers:
Should I rewrite that part of the README given the code is no longer in the noria crate? I would propose moving that line including the `controller/inner.rs` to the server section of the README.

ms705:
Thanks! The README is fine as-is, since it refers to the client _invoking_ controller RPCs.

PR #170, opened 2020-11-28
Fix a couple unnecessary mutable warns

glittershark:
Fix a couple "variable does not need to be mutable" warnings by dropping
a couple `mut`s


ms705:
Thank you! :+1:

PR #174, opened 2021-02-07
chore: use stable

Nonius:
I couldn't find any documentation why use the nightly compiler
but since it compiles and tests pass it might be nice to use stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: trusty
 language: rust
 cache: false
 rust:
-  - nightly
+  - stable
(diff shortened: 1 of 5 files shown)


Nonius:
I think that the pipeline should pass if both this and #176 are merged.

Nonius:
I also wonder why does`cargo check --workspace --all-features --all-targets` works on my machine with `stable`.

Nonius:
Since I used `RUSTC_BOOTSTRAP=1` this request makes no sense.

